PR #12, opened 2015-04-27
Pass token in from the blueprint backend.

pdex:
The quickstart examples are failing because the OAuth2Session authorized method is always returning False. I tracked it down and it turns out that the token isn't being loaded from the blueprint backend. I'm not sure if this is the best way to fix this, but it works for me.


singingwolfboy:
@pdex: Thanks for reporting this issue. I put in a proper fix in #13, and I release Flask-Dance 0.5.1 on PyPI. Can you test with that version, and see if the quickstart is working for you now?

pdex:
#13 works for me, thanks for your fast response.

PR #13, opened 2015-04-28
Make the `authorized` property load the token from the backend

singingwolfboy:
To properly fix the bug identified in #12. Unfortunately, the `OAuth1Session` tests fail on the current version of requests-oauthlib (0.4.2), because it still contains the bug fixed by https://github.com/requests/requests-oauthlib/commit/b94f16472d66f5ed0febb8c1985854f379f7bac3.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+unreleased
+----------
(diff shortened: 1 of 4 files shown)


pdex:
AWESOME, I'll test this out tonight!

PR #16, opened 2015-05-04
Make sure the redirect URI is always set

Bachmann1234:
Some providers like facebook require the request id used to authorize the app when making authenticated requests.

In my debugging I was getting back that the redirect_uri was not there. If we don't hit login this property is not set. There may be a better place to do it but I thought id send over what I had to do

### DIFF
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -154,12 +154,15 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
     def teardown_session(self, exception=None):
         lazy.invalidate(self, "session")
 
-    def login(self):
+    def set_redirect_uri(self):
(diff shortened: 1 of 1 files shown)


Bachmann1234:
In case you are curious here is the facebook configuration I got working

https://gist.github.com/Bachmann1234/fec472272ce79bb707d7

Im hoping to turn it into a blueprint at somepoint. Maybe this week. However, if someone else feels like doing it or needs to get facebook working feel free :-)

singingwolfboy:
I believe that this change is not necessary. Check the latest version on master -- [it sets the `redirect_uri` in the `authorized` view](https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/consumer/oauth2.py#L199-L203). (Not yet released on PyPI.)

Bachmann1234:
Ah, yes. I think you warned me about this actually.... Cool

PR #17, opened 2015-05-06
Bubble up the params that caused a MissingTokenError

Bachmann1234:
When experimenting with Facebook I became frustrated because I misconfigured my app
screen shot 2015-05-06 at 12 39 41 am

Had I been observant I would have noticed the error message in the URL. But I was looking at flasks debug message which was useless. So I ended up digging into the source. Not to mention I am not fluent in URL encoding

To prevent me or others from needing to do this again I wrapped the error up in a way that displays in the browser stack trace

screen shot 2015-05-06 at 12 39 05 am

### DIFF
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -1,9 +1,11 @@
 from __future__ import unicode_literals, print_function
+import json
 
 import logging
 from lazy import lazy
 import flask
 from flask import request, url_for, redirect
+from oauthlib.oauth2 import MissingCodeError, MissingTokenError, OAuth2Error
 from urlobject import URLObject
 from .base import (
     BaseOAuthConsumerBlueprint, oauth_authorized, oauth_error
@@ -205,12 +207,19 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
         url = URLObject(request.url)
         if request.headers.get("X-Forwarded-Proto", "http") == "https":
             url = url.with_scheme("https")
-        token = self.session.fetch_token(
-            self.token_url,
-            authorization_response=url,
-            client_secret=self.client_secret,
-            **self.token_url_params
-        )
+        try:
+            token = self.session.fetch_token(
+                self.token_url,
+                authorization_response=url,
+                client_secret=self.client_secret,
+                **self.token_url_params
+            )
+        except MissingCodeError:
+            raise MissingCodeError(
+                "The redirect request did not contain the expected parameters. Instead I got: {}".format(
+                    json.dumps(request.args)
+                )
+            )
         results = oauth_authorized.send(self, token=token) or []
         if not any(ret == False for func, ret in results):
             self.token = token
--- a/tests/consumer/test_oauth2.py
+++ b/tests/consumer/test_oauth2.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, unicode_literals
+from oauthlib.oauth2 import MissingTokenError, MissingCodeError, OAuth2Error
 
 try:
     from urllib.parse import quote_plus, parse_qsl
@@ -23,7 +24,7 @@ except ImportError:
 requires_blinker = pytest.mark.skipif(not blinker, reason="requires blinker")
 
 
-def make_app(login_url=None):
+def make_app(login_url=None, debug=False):
     blueprint = OAuth2ConsumerBlueprint("test-service", __name__,
         client_id="client_id",
         client_secret="client_secret",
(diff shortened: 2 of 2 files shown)


flask_dance/consumer/oauth2.py
```diff
@@ -205,12 +207,19 @@ def authorized(self):
...
-        )
+        try:
+            token = self.session.fetch_token(
+                self.token_url,
+                authorization_response=url,
+                client_secret=self.client_secret,
+                **self.token_url_params
+            )
+        except MissingCodeError:
+            raise MissingCodeError(
```
singingwolfboy: Can you attach the original exception to the new exception that you're raising? That way, it's possible for code to introspect the new exception and figure out what the original exception was.

flask_dance/consumer/oauth2.py
```diff
@@ -1,9 +1,11 @@
 from __future__ import unicode_literals, print_function
+import json
 
 import logging
 from lazy import lazy
 import flask
 from flask import request, url_for, redirect
+from oauthlib.oauth2 import MissingCodeError, MissingTokenError, OAuth2Error
```
singingwolfboy: Looks like you don't need all these imports.

tests/consumer/test_oauth2.py
```diff
@@ -127,6 +129,24 @@ def test_authorized_url():
...
+        with client.session_transaction() as sess:
+            sess["test-service_oauth_state"] = "random-string"
+        # make the request
+        with pytest.raises(MissingCodeError) as missingError:
+            client.get(
+                "/login/test-service/authorized?error_code=1349048&error_message=IMUSEFUL",
+                base_url="https://a.b.c",
+            )
+        assert str(missingError.value) == ('(missing_code) The redirect request did not contain a token. '
+                                           'Instead I got: {"error_message": "IMUSEFUL", "error_code": "1349048"}')
```
singingwolfboy: When a dict is coerced to a string, the order of the key-value pairs is undefined. As a result, this test could fail on a different Python implementation that orders the key-value pairs differently. You could either do:
```python
assert str(missingError.value) in ('ordering 1', 'ordering 2')
```
or you can parse `missingError.value` back into a dict, and compare it with an actual dict.

Bachmann1234:
This is what I get for a late night PR. My bad. Ill clean this up and fix my own test tonight.... Sorry!

tests/consumer/test_oauth2.py
```diff
@@ -127,6 +129,24 @@ def test_authorized_url():
...
+
+
+
+
```
singingwolfboy: Extra whitespace here. I need to set up codeclimate or something.

tests/consumer/test_oauth2.py
Bachmann1234: Parsing the dict may be difficult.

ill poke at this some. I may do the approach of checking for key words.

For the purpose of the test I just wanna make sure each word is in there

```
assert all(keyword in missingError.value for keyword in ["error_message",  "IMUSEFUL", "error_code",  "1349048"]
```

but when I am with the IDE tonight ill look for other solutions

flask_dance/consumer/oauth2.py
Bachmann1234: No... no I do not.

flask_dance/consumer/oauth2.py
Bachmann1234: ooo, good call!

tests/consumer/test_oauth2.py
Bachmann1234: probably a good call but I should have run a tool on my end.

tests/consumer/test_oauth2.py
singingwolfboy: If I were you, I'd just use a regex to pull out the dict:
```python
import re
match = re.search("{[^}]*}", missingError.value)
err_dict = json.loads(match.group(0))
assert err_dict == {"error_message": "IMUSEFUL", "error_code": "1349048"}
```

tests/consumer/test_oauth2.py
Bachmann1234: Another option would be to add diff-quality to your build #alwaysBeSelling

Set fail-under and you can see the nice little X 

Less polished by lighter weight!

Bachmann1234:
PR is embarrassing me so i'm closing and will reopen tonight

singingwolfboy:
Dude, your PRs are awesome. Don't let my nitpicks bring you down. :heart:

Bachmann1234:
@singingwolfboy I wanna be clear. I'm closing because the PR fails a test (ITS OWN TEST) and has obvious flake8 problems. Which I find inappropriate. 

Love the feedback on the exception handling and improving test. If that was it I would leave the PR up. Promise this will come back up soon

Issue #15, opened 2015-05-04
Consider bubbling oauth providers up


Bachmann1234:
I can look into a PR for this if you like the idea....

So I have been playing with using flask-dance with facebook oauth (potential PR there as well)

Facebook has an... lets call it advanced... control panel for its apps. For a newbie it can be easy to make mistakes. Personally I found it difficult to debug these mistakes when using flask-dance.

So take redirect uri that's totally wrong! When I test my app I see this

![screen shot 2015-05-03 at 10 35 03 pm](https://cloud.githubusercontent.com/assets/1521093/7448344/4ac21368-f1e5-11e4-8082-fc01ff15e09a.png)

But thats absolutely useless to me. Dig into the source and there is a more useful error message lurking

![screen shot 2015-05-03 at 10 44 57 pm](https://cloud.githubusercontent.com/assets/1521093/7448373/11f02fc4-f1e6-11e4-84de-862f4bf5794c.png)

```'{"error":{"message":"Error validating verification code. Please make sure your redirect_uri is identical to the one you used in the OAuth dialog request","type":"OAuthException","code":100}}'```

Now I know to fix my facebook app settings

![screen shot 2015-05-03 at 10 33 31 pm](https://cloud.githubusercontent.com/assets/1521093/7448382/2e0aca7a-f1e6-11e4-9b21-9bcfdf17349d.png)

Perhaps you can surface the response text when a request fails.

singingwolfboy:
Definitely a good idea. Maybe we can put a try/except block around `self.session.fetch_token`, and have it throw a more useful exception with the response from the provider.

Bachmann1234:
Yeah, sounds good. But im tired of code tonight. Ill make a note to address this sometime soon

singingwolfboy:
Totally reasonable. Thank you so much for all your help on this! :sparkling_heart:

PR #18, opened 2015-05-06
Bubble up the params that caused a MissingTokenError

Bachmann1234:
When experimenting with Facebook I became frustrated because I misconfigured my app
screen shot 2015-05-06 at 12 39 41 am

Had I been observant I would have noticed the error message in the URL. But I was looking at flasks debug message which was useless. So I ended up digging into the source. Not to mention I am not fluent in URL encoding

To prevent me or others from needing to do this again I wrapped the error up in a way that displays in the browser stack trace

screen shot 2015-05-06 at 7 30 09 pm

I addressed the comments and obvious sloppiness from the previous PR. Specifically the test should work with any complaint python interpreter and I modify and reraise the original exception rather than making a new one

*im assuming that you still want python 2 support so I did not use exception chaining*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ unreleased
 * Added the Meetup pre-set configuration
 * Flask-Dance now always passes the optional ``redirect_uri`` parameter to
   the OAuth 2 authorization request, since Dropbox requires it.
+* Make Flask-Dance provide additional information in errors when providers fail
+  to provide auth tokens
 
 0.5.1 (2015-04-28)
 ------------------
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -1,9 +1,11 @@
 from __future__ import unicode_literals, print_function
+import json
 
 import logging
 from lazy import lazy
 import flask
 from flask import request, url_for, redirect
+from oauthlib.oauth2 import MissingCodeError
 from urlobject import URLObject
 from .base import (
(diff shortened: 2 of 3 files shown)


Bachmann1234:
This should address https://github.com/singingwolfboy/flask-dance/issues/15

singingwolfboy:
This looks great! I've turned off the annoying coveralls comments. Can you add an entry to the `CHANGELOG.rst` file for this change, as well?

singingwolfboy:
Also, can you make your automated test check that the original exception information is still accessible? (I'm not sure what the best way of doing that is -- if the answer is "no, I can't do that", then that's an acceptable answer and won't block a merge. Just raising the idea.)

Bachmann1234:
I added the CHANGELOG.

Tell me a bit more about this test update you are thinking about.

I could check if the original error string is there

```assert "Missing Code parameter in response" in str(missingError.value)```

Or did you mean ensure the stack trace is in tact...

I could check to make sure the line where the exception is raised is still there.... but I worry this will be fragile if the underlying library changes.

Something like....

```
assert 'parse_authorization_code_response' in [entry.name for entry in missingError.traceback]
```

singingwolfboy:
Oh, you replaced the exception message but kept the original stack trace. I was thinking that you would attach the original exception to the new exception as an attribute, in which case you could check that the attribute was there -- but this works as well! You're right, though, I think this is too fragile to test. I think this is ready to go. :+1:

PR #29, opened 2015-09-17
Added Bitbucket provider

jsfan:
Bitbucket requires HTTP Basic Authentication witch client_id and client_secret to fetch tokens. `OAuth2ConsumerBlueprint` has been adjusted to pass through authentication details (object or tuple as expected by `requests`).

A provider `Bitbucket` has been added.

Test for new provider has been added.

Documentation has been updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,6 +9,7 @@ README will give you a taste of the features.
 Flask-Dance currently provides pre-set OAuth configurations for the following
 popular websites:
 
+* Bitbucket
 * Facebook
 * GitHub
 * Google
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -11,6 +11,18 @@ to Flask-Dance!
    :local:
    :backlinks: none
 
+Bitbucket
+--------
+.. module:: flask_dance.contrib.bitbucket
+
+.. autofunction:: make_bitbucket_blueprint
+
+.. data:: bitbucket
+
+    A :class:`~werkzeug.local.LocalProxy` to a :class:`requests.Session` that
+    already has the Bitbucket authentication token loaded (assuming that the user
+    has authenticated with Bitbucket at some point in the past).
+
 Facebook
 --------
 .. module:: flask_dance.contrib.facebook
--- a/docs/quickstarts/bitbucket.rst
+++ b/docs/quickstarts/bitbucket.rst
@@ -1,0 +1,84 @@
+Bitbucket Quickstart
+=================
+
+Setup Application
+-----------------
+Visit https://bitbucket.org/account/user/nothingbutweb/api
+to register an application on Bitbucket. The application's "authorization
+callback URL" must be ``http://localhost:5000/login/bitbucket/authorized``.
+Take note of the "Client ID" and "Client Secret" for the application.
+
+Code
+----
+.. code-block:: python
+
+    from flask import Flask, redirect, url_for
+    from flask_dance.contrib.bitbucket import make_bitbucket_blueprint, bitbucket
+
+    app = Flask(__name__)
+    app.secret_key = "supersekrit"
+    blueprint = make_bitbucket_blueprint(
+        client_id="my-key-here",
+        client_secret="my-secret-here",
+    )
+    app.register_blueprint(blueprint, url_prefix="/login")
+
+    @app.route("/")
+    def index():
+        if not bitbucket.authorized:
+            return redirect(url_for("bitbucket.login"))
+        resp = bitbucket.get("/api/1.0/user")
+        assert resp.ok
+        return "You are @{login} on Bitbucket".format(login=resp.json()["user"]["username"])
+
+    if __name__ == "__main__":
+        app.run()
+
+.. note::
+    You must replace ``my-key-here`` and ``my-secret-here`` with the client ID
+    and client secret that you got from your Bitbucket application.
+
+.. note::
+    If you are running this code on Heroku, you'll need to use the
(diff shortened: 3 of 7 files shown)


singingwolfboy:
Hi @jsfan, thanks for the pull request! However, can you explain the new `auth` parameter you're creating, and why it's needed? One of the main reasons why OAuth exists in the first place is to get away from passing username/password combinations over the internet like this.

jsfan:
Unfortunately, your statement is incorrect. OAuth was never designed to handle _authentication_ but _authorisation_ only. Authentication is still addressed in RFC 6749, however, and the most common authentication of implicitly authenticating as part of the token request is actually discouraged there.

If you have a look at [Section 2.3.1 of RFC 6749](https://tools.ietf.org/html/rfc6749#section-2.3.1), you will notice that [Atlassian's implementation](https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-238027431.html#OAuthonBitbucket-Granttypes) of OAuth2 is actually by the book while that of other providers implements the less preferred option.

To be honest, I only implemented the authentication passthrough because Bitbucket required it. But, having done the research (prompted by your challenge), I tip my hat to Atlassian to following the RFC more closely that most other providers.

singingwolfboy:
Interesting -- I didn't realize that authenticating with a username/password combination was not only allowed by the RFC, but *encouraged*. That's good to know. However, I don't believe that Bitbucket *requires* username/password authentication -- I don't see anything about that in the documentation, and [this example written by Bitbucket](https://confluence.atlassian.com/bitbucket/oauth-consumer-examples-338367592.html) doesn't ask for a username or password at all.

I'd prefer to handle separate concerns in separate pull requests, rather than lumping multiple changes into one; that makes the pull requests easier to review and merge. As a result, I suggest that this pull request focus on merging the Bitbucket provider into Flask-Dance *without* username/password support, and then making a separate pull request that provides username/password support throughout Flask-Dance, so that the Bitbucket provider can use it. Does that make sense?

jsfan:
The sample code you found uses OAuth1 the use of which Atlassian discourages. In that case authentication is handled by simply requiring the use to have a Bitbucket session. I don't think you really want an OAuth1 implementation for Bitbucket in Flask Dance under these circumstances.

All of Atlassian's documentation regarding OAuth2 requires authentication with the HTTP Auth as you can see from the `curl` calls on the page I linked to above.

I am happy to split the concerns off into two pull requests. However, if I stick with OAuth2, my Bitbucket provider will always require the first commit which passes through the authentication details.

How would you like me to proceed?

singingwolfboy:
I really don't think that HTTP Basic auth is necessary. I've put together an example app at https://github.com/singingwolfboy/flask-dance-bitbucket, based on the code I already had at https://github.com/singingwolfboy/flask-dance-github. I'm running it at https://flask-dance-bitbucket.herokuapp.com, and it's not yet working, but I'm not sure the failure is due to a lack of HTTP Basic auth -- I've able to get to BitBucket's authorization screen, at least. I'll debug it more later, I just wanted to let you know that I'm fiddling around with this.

jsfan:
I suspect that you are getting as far I do. I actually set up OAuth2 with a custom provider at first and got a "400 Bad Request". Only when I stepped through my flask app on my development server did I find out that the authentication was the problem.

If you want to debug it, break at line 212 in oauth2.py and see if fetch_token throws an exception. I believe if you have logging enabled, requests_oauthlib will also log what exactly happened. Unfortunately, the details as provided by Bitbucket are never passed through from requests_oauthlib and will not be shown on the Flask development server even in debug mode.

jsfan:
Just as a word of warning, I have found out that you will run into a new problem regarding authentication when your token has expired (Bitbucket tokens are only valid for 1h). requests_oauthlib doesn't pass through authentication details on refresh requests despite implementing the interface internally. I will make a pull request against request_oauthlib as well.

I also noticed that I am not setting the refresh URL in my provider which will always result in a TokenExpired exception independent of authentication working or not. I will update my pull request.

jsfan:
I've noticed that there are still more issues with refreshing tokens. While requests-oauthlib now passes on the necessary authentication details, flask-dance requires the authentication details to be sent along with the API call. This probably needs to happen automatically within flask-dance to make it a true auto refresh.

Maybe, leave this pull request for now until I have had a chance to work out exactly what still needs to happen for it to work smoothly.

singingwolfboy:
@jsfan Makes sense. Thanks for all your work on this -- having more people involved in the project is great! I built Flask-Dance to scratch my own itch, so it works better for the things that I needed, and anything that I didn't need might not work as well, so I'm not surprised that refreshing tokens has some kinks to work out.

Would you like me to close this pull request, and you can open a new one when you've got things working the way you want? Or should I leave it open? Up to you.

jsfan:
I'm sorry I never answered the question. I did not succeed when trying to patch requests-oauthlib and have not be able to revisit it since. I'll make a new pull request as soon as I have the requests-oauthlib part working. Too many projects, too little time...

singingwolfboy:
I think we're going to need to go all the way down to oauthlib for this, actually. Would you like to find a time to paircode on this together, maybe through Google Hangouts or something?

singingwolfboy:
@jsfan: it looks like this change has happened! Check out #47.

jsfan:
That's great news. The project that made me stumble upon this issue has been history for a while, so I didn't really get any chance to look into the [requests-oauthlib](https://github.com/requests/requests-oauthlib) part of this issue.

Not sure if a fix on their end will just make my provider work. I suspect there will still be need to adapt. I'll keep it on my todo list. Can't really promise any more than that at the moment.

PR #36, opened 2015-11-29
If missing "state" variable, redirect to `login` view

### LINKED ISSUE #34: KeyError: 'google_oauth_state'
Hello,

I receive the error `KeyError: 'google_oauth_state'` in case the browser back button is hit after successful authentication (at least in the google authentication example provided)

### How to reproduce
Take this script which basically resembles the google authentication example
~~~~.python
from flask import Flask, redirect, url_for
from flask_dance.contrib.google import make_google_blueprint, google

app = Flask(__name__)
app.secret_key = "supersekrit"
app.debug = True
import os
os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = "1"

blueprint = make_google_blueprint(
    client_id="my-key-here",
    client_secret="my-secret-here",
    scope=["profile", "email"]
)
app.register_blueprint(blueprint, url_prefix="/login")

@app.route("/")
def index():
    if not google.authorized:
        return redirect(url_for("google.login"))
    resp = google.get("https://www.googleapis.com/oauth2/v1/userinfo")
    assert resp.ok, resp.text
    return resp.text

if __name__ == "__main__":
    app.run()
~~~~
(Note that I set the OAUTHLIB_INSECURE_TRANSPORT environment variable as described in the docs but then in python code. Furthermore I used a different info url and just dump the whole response text. But I don't think that is relevant here)

The point is, if you start the app and navigate to it in the browser, You are sent to the google login screen. After that, you will be redirected to the appropriate blueprint url which then points you back to the index route. Fetching the userinfo is no problem either.
But when you hit the back button in your browser, the `KeyError` occurs which originates at flask_dance/consumer/oauth2.py on line 196.

singingwolfboy:
If the user hits the ``authorized`` view without having a "state" variable set in the browser cookies, Flask-Dance will now redirect the user back to the ``login`` view to start the OAuth dance all over again, rather than raising a ``KeyError``. Fixes #34.

@rmoorman, can you take a look at this, and test it before it gets merged?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,10 @@ unreleased
 * Removed the Dictective utility class, and replaced it with
   ``werkzeug.datastructures.CallbackDict``. It does the same thing, but
   it's better tested, and already a part of one of Flask-Dance's dependencies.
+* If the user hits the ``authorized`` view without having a "state" variable
+  set in the browser cookies, Flask-Dance will now redirect the user back
+  to the ``login`` view to start the OAuth dance all over again, rather than
+  raising a ``KeyError``.
 
 0.7.0 (2015-08-21)
 ------------------
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -192,6 +192,10 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             return redirect(next_url)
 
         state_key = "{bp.name}_oauth_state".format(bp=self)
+        if state_key not in flask.session:
+            # can't validate state, so redirect back to login view
+            return redirect(url_for(".login"))
+
         self.session._state = flask.session[state_key]
         del flask.session[state_key]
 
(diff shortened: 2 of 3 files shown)


rmoorman:
@singingwolfboy looks great. I will test it and report if it works as expected on my side too.

singingwolfboy:
Great, thank you!

singingwolfboy:
@rmoorman: Have you gotten a chance to test this yet? If it works, I'd love to merge it.

rmoorman:
@singingwolfboy so far it is working ok. In case the back button is hit, I get redirected to the permission screen again. Only one thing is a bit strange.
Right now, I get the permission screen, after that I can choose the google identity and after that I have to accept the permissions again. :-)

I will try to have a look if I can find where this is behavior coming from.

singingwolfboy:
Alright, that sounds like a separate issue. I'll merge this pull request, and cut a new Flask-Dance release either tonight or in the next few days.

rmoorman:
great!

PR #45, opened 2015-12-30
Redirect on OAuth 1 login error

singingwolfboy:


singingwolfboy:
coverage drop is expected, and small enough that I don't really care.

PR #52, opened 2016-03-12
call to fetch access token should be signed with request token secret

kellan:
Hi.

I'm completely new to flask (and flask_dance) and pretty rusty on Python.

But I noticed that the flask_dance oauth1.0 consumer doesn't use the request token secret to sign the request for the access token.  Twitter doesn't care, but most 1.0a implementation should.

I store the request_token in the flask's session during login so it's available during authorize

### DIFF
--- a/flask_dance/consumer/oauth1.py
+++ b/flask_dance/consumer/oauth1.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals, print_function
 import logging
 from lazy import lazy
 from flask import request, url_for, redirect
+from flask import session as flask_session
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Hi @kellan, can you provide a concrete example of what you're trying to accomplish with this change? Which 1.0a implementation are you testing with? Even better, can you add an automated test that demonstrates the issue you're trying to fix?

singingwolfboy:
@kellan, do you know when you'll be able to respond to my comment? If I don't hear back from you, I can't review your pull request, and I'll have to close it.

PR #55, opened 2016-05-18
Fix docs for custom oauth consumer

ThiefMaster:


singingwolfboy:
D'oh! Thanks for fixing this. 👍

PR #56, opened 2016-05-18
`pkg_resources.parse_version()` instead of `distutils.StrictVersion`

### LINKED ISSUE #53: StrictVersion check fails with Flask==0.11.dev0
```
  File "/home/.../.venv/lib64/python3.5/site-packages/flask_dance/consumer/base.py", line 37, in __init__
    if StrictVersion(flask.__version__) < StrictVersion('1.0'):
  File "/usr/lib64/python3.5/distutils/version.py", line 40, in __init__
    self.parse(vstring)
  File "/usr/lib64/python3.5/distutils/version.py", line 137, in parse
    raise ValueError("invalid version number '%s'" % vstring)
ValueError: invalid version number '0.11.dev0'
```

singingwolfboy:
`pkg_resources.parse_version()` is more flexible. Fixes #53.

@ThiefMaster, can you try this out and let me know if it works for you?

### DIFF
--- a/flask_dance/consumer/base.py
+++ b/flask_dance/consumer/base.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals, print_function
 import six
 from lazy import lazy
 from abc import ABCMeta, abstractmethod, abstractproperty
-from distutils.version import StrictVersion
+from pkg_resources import parse_version
 from werkzeug.datastructures import CallbackDict
 import flask
 from flask.signals import Namespace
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Haven't tested it yet but I think Flask>=0.11 should be accepted since 0.11 is current master (which will be released as 1.0 eventually), so it's enough. I don't use `root_path` but no need to remove it in a version where it's supported. Or why not pass it only to the blueprint constructor if it's not None? That way setting it will fail if Flask is too old and work fine if not. I think that's better than silently ignoring the argument if it's not supported.

singingwolfboy:
@ThiefMaster: good thought, I've updated the pull request.

ThiefMaster:
thanks, works fine now!
could you release 0.8.3 by any chance after merging this? would be nice not having to switch to a git version to get the fix (right now i'm using a monkeypatch but that's awful and i'd rather not commit that if a new version fixes it properly)

singingwolfboy:
I've released 0.8.3 with the fix.

ThiefMaster:
great, thanks! if all projects/maintainers were that fast... ;)

PR #62, opened 2016-08-27
Improvements for OAuth2 auto refresh

singingwolfboy:
While testing Flask-Dance with Bitbucket, I found that the OAuth2 auto token refresh wasn't working properly. I believe this makes it work right.

PR #64, opened 2016-09-17
Added the Azure AD pre-set configuration

steven-martins:
Hello,

I currently have project which needs an authentication with Azure AD and the v2.0 endpoint. So, I decided to create a new pre-set.

Regards,

Steven.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,10 @@ unreleased
 ----------
 Nothing yet.
 
+0.9.1 (2016-09-17)
+-----------------
+* Added the Azure AD pre-set configuration
+
 0.9.0 (2016-07-1)
 -----------------
 * Allowed an ``oauth_authorized`` event handler to return a ``flask.Response``
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,6 @@ Change Log
 
 unreleased
 ----------
-Nothing yet.
-
-0.9.1 (2016-09-17)
------------------
 * Added the Azure AD pre-set configuration
 
 0.9.0 (2016-07-1)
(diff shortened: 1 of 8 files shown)


singingwolfboy:
This looks great! However, can you take out the changes related to the version number, and put the changelog entry under "unreleased"? After this gets merged, I'll do the changes necessary for a new release of the project.

singingwolfboy:
Also, what is "Azure AD", and how is it different from Azure in general? It looks to me like the code you've put in should work for any Azure service in general, but I'm not familiar with Azure.

steven-martins:
I pushed a new commit to remove the version number.

AD is for "Active Directory", it's their authentication system / Users database. Every documentations of Microsoft append 	specifically "AD" when they talk about Single sign-on, oauth, etc. ... According to me "Azure" is more oriented to their cloud platform.

If you want, I can remove any mention of "AD" ?

singingwolfboy:
Sorry for the delay -- I was on my phone when I first saw this pull request. Now that I've had some time to examine it more in-depth, and review Azure's documentation, I think this is all set. You're right that Microsoft refers to it as "Azure AD" everywhere in their documentation, so it makes sense to have the Flask-Dance documentation reflect that.

Thank you so much for the contribution! ✨ 🍰 ✨

singingwolfboy:
I'll release a new version of Flask-Dance to PyPI in a few days -- I want to see if we can also get #62 in, as well. That pull request is waiting on a newly-released version of `requests-oauthlib`, which [the maintainer said he would do in a few days](https://github.com/requests/requests-oauthlib/pull/251#issuecomment-247850377). Once that's in, I'll cut and release Flask-Dance 0.10!

singingwolfboy:
[Flask-Dance 0.10.0 is out!](https://pypi.python.org/pypi/Flask-Dance/0.10.0)

PR #65, opened 2016-11-21
add auto_refresh_url to oauth blueprint creation

martinffx:
The google blueprint is not setting the auto_refresh_token_url. So the token is not automatically refreshing when it has expired and a valid refresh token exists.

As per: https://github.com/requests/requests-oauthlib/blob/master/docs/oauth2_workflow.rst#third-recommended-define-automatic-token-refresh-and-update

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,8 @@ Change Log
 
 unreleased
 ----------
-nothing yet
+* Fixed ``make_google_blueprint`` to include ``auto_refresh_url`` so that
+  token renewal is automatically handled by ``requests-oauthlib``
 
 0.10.0 (2016-09-27)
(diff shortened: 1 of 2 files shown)


singingwolfboy:
Can you add an item to the `CHANGELOG` file, under the "Unreleased" section, to account for this change?

martinffx:
When do you think you'll put together another release and push to pypi?

singingwolfboy:
@martinffx I just tagged 0.10.1 and pushed it to PyPI. 😄

martinffx:
@singingwolfboy Thanks! And thanks for the awesome library.

singingwolfboy:
You're very welcome! Thanks for helping to improve it! 💖

PR #71, opened 2017-02-16
Drop sqlalchemy_utils

singingwolfboy:
Now that [SQLAlchemy has JSON support built-in in version 1.1 and higher](http://docs.sqlalchemy.org/en/latest/changelog/migration_11.html#json-support-added-to-core), we should depend on SQLAlchemy 1.1 and drop `sqlalchemy_utils`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,9 @@ Change Log
 
 unreleased
 ----------
-nothing yet
+* Depend on SQLAlchemy 1.1 or higher, so that we can drop the dependency on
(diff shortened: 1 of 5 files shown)


singingwolfboy:
Hmm, sqlite doesn't work with `sqlalchemy.types.JSON`. That might make this a no-go...

wgwz:
Not sure if this helps at all but a quick google search "sqlite json types" yielded this result: https://sqlite.org/json1.html

If this JSON support is happening at the SQLA level, shouldn't it be handled internally there? i.e. if you are using a sqlite database, throw an error if trying to use a JSON type. 

Perhaps I misunderstand entirely :D

PR #72, opened 2017-04-07
timezone handling in timestamp_from_datetime

aongko:
My machine was set on UTC+7.
When the **value** of `dt` in `timestamp_from_datetime(dt)` is in UTC, such as in https://github.com/singingwolfboy/flask-dance/blob/deab6261c68e7eb643b0c174a9f4f0dd873d8316/flask_dance/consumer/base.py#L113
The function will treat `dt` as it is in UTC+7.
For example, when `dt = datetime.utcnow()` and `dt` becomes "2017-01-01 00:00:00.000", `dt.timestamp()` will convert `dt` to UNIX timestamp while assuming "2017-01-01 00:00:00.000" as UTC+7

### DIFF
--- a/flask_dance/utils.py
+++ b/flask_dance/utils.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 import functools
 from collections import MutableMapping
-from datetime import datetime
+from datetime import datetime, timezone
 
 
 class FakeCache(object):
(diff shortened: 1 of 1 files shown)


aongko:
@singingwolfboy care to review this?

singingwolfboy:
Thanks for catching this! The automated tests are failing, but that doesn't seem to be the fault of your pull request; it appears that one of the dependencies updated and revealed a problem in the code. I fixed that in #73, so I'm going to merge this anyway, under the assumption that the problem that caused the test failures is already resolved.

PR #85, opened 2017-10-27
Update backends.rst

NathanWailes:
change "database" to the more-technically-correct "ORM", with a link to a simple Stack Overflow explanation of what an ORM is.

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -115,4 +115,4 @@ backend that uses a file on disk::
 Then, just create an instance of your backend and assign it to the
 :attr:`backend` attribute of your blueprint, and Flask-Dance will use it.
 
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Thanks! However, I feel weird about linking to Stack Overflow for a definition. Stack Overflow is a useful tool, but it's not always reliably correct.

Can you find a website that has a more authoritative definition of what an ORM is, and link to that? Or maybe just include a brief definition in the `backends.rst` file directly?

NathanWailes:
Switched to [the official Python docs](https://docs.python.org/3.4/howto/webservers.html#data-persistence)

singingwolfboy:
That works much better. Thanks!

PR #86, opened 2017-10-28
Update how-oauth-works.rst

NathanWailes:
Fixed the numbering for the OAuth 1 explanation


singingwolfboy:
Good catch! Thanks for the fix. ❤️

PR #91, opened 2017-10-31
Add to the documentation

NathanWailes:
Here are just some ideas for comments that I think might make it easier for future users to understand how everything works. I'm open to changing things (e.g. moving explanations into the docs, or changing the way things are explained). But I think those functions really could benefit from being explained in-depth.


singingwolfboy:
Thanks!

Issue #88, opened 2017-10-29
How does twitter.authorized know it's me?


NathanWailes:
I don't really understand how I can log out, delete my cookie, then click the "Sign in with Twitter" button and it somehow knows who I am and just logs me in directly. `twitter.authorized` is evaluating to `True`, and I saw that the code I think that's being evaluating to True is:

    bool(self._client.client.client_secret) and
    bool(self._client.client.resource_owner_key) and
    bool(self._client.client.resource_owner_secret)

How can the session still have a `resource_owner_key` and `resource_owner_secret` after I've logged out, deleted my session cookie in Chrome, and restarted the server?

singingwolfboy:
Because although you've logged out of your local website, you _haven't_ logged out of Twitter.com. When you try to log in with Twitter using OAuth, the following things happen:

1. Flask-Dance sends you to Twitter, along with some information about who sent you there.
2. Twitter checks who sent you, and asks that you log in to Twitter. After you're logged in, Twitter asks you if you're OK with sharing your information with the website that sent you.
3. When you agree, Twitter sends you back to the original website, along with a special token that the original website can use to call the Twitter API on your behalf.

Even if you log out of your website, restart everything, and so on, that doesn't change anything from Twitter's point of view. When you get to step 2 of the OAuth dance, Twitter is smart enough to see that you are _already_ logged in on Twitter, and you have _already_ agreed to sharing your information with this particular website, so they don't bother to make you confirm a second time. Instead, they just send you right back to the original website, with the special token for that website to use.

Does that make sense? If you want to truly start over the OAuth dance (usually for purposes of testing), you have to go to the provider website and revoke your permission for sharing information with the consumer website. In the case of Twitter, specifically, go to [the Apps section of your Twitter settings](https://twitter.com/settings/applications) and revoke the permission for your Twitter app. After you do that, if you try to use Flask-Dance to login with Twitter, Twitter will once again ask you for permission to share your information with the other website.

NathanWailes:
Thank you for that in-depth explanation!

So is it accurate to say that calling `twitter.authorized` pings Twitter?

singingwolfboy:
No, it doesn't. If you have code that looks like this:

```python
@app.route('/')
def index():
    if not twitter.authorized:
        return redirect(url_for('twitter.login'))
    # view continues here
```

The first time you hit this code, `twitter.authorized` will be False. That will start a chain of redirects that:

* sends you to Twitter
[...]

PR #92, opened 2017-11-04
Make sqla backend know when to require a user

singingwolfboy:
This resolves the issue raised in #88, where Flask-Dance should have raised an exception instead of trying to create an OAuth token without an associated user. This changes the SQLAlchemy backend to take a new optional argument: `require_user`. When set to True, the backend will not allow OAuth tokens to be created without an associated user. This argument is True by default when an argument is passed for `user` or `user_id`.

@NathanWailes, can you take a look at this, and let me know if the functionality is what you had in mind?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,9 @@ unreleased
   documentation.
 * Added Spotify pre-set configuration
 * Added Discord pre-set configuration
+* Added an optional ``user_required`` argument to the SQLAlchemy backend.
+  When this is enabled, trying to set an OAuth object without an associated
+  user will raise an error.
 
 0.12.0 (2017-10-22)
 -------------------
--- a/flask_dance/consumer/backend/sqla.py
+++ b/flask_dance/consumer/backend/sqla.py
@@ -57,7 +57,8 @@ class SQLAlchemyBackend(BaseBackend):
     .. _SQLAlchemy: http://www.sqlalchemy.org/
     """
     def __init__(self, model, session,
-                 user=None, user_id=None, anon_user=None, cache=None):
+                 user=None, user_id=None, user_required=None, anon_user=None,
+                 cache=None):
         """
         Args:
             model: The SQLAlchemy model class that represents the OAuth token
@@ -84,6 +85,13 @@ class SQLAlchemyBackend(BaseBackend):
                 User object, use this argument instead. Sometimes it can save
                 a database query or two. If both ``user`` and ``user_id`` are
                 provided, ``user_id`` will take precendence.
+            user_required:
+                If set to ``True``, an exception will be raised if you try to
+                set or retrieve an OAuth token without an associated user.
+                If set to ``False``, OAuth tokens can be set with or without
+                an associated user. The default is auto-detection: it will
+                be ``True`` if you pass a ``user`` or ``user_id`` parameter,
+                ``False`` otherwise.
             anon_user:
                 If anonymous users are represented by a class in your application,
                 provide that class here. If you are using `Flask-Login`_,
@@ -102,6 +110,10 @@ class SQLAlchemyBackend(BaseBackend):
         self.session = session
         self.user = user
         self.user_id = user_id
+        if user_required is None:
+            self.user_required = user is not None or user_id is not None
+        else:
+            self.user_required = user_required
         self.anon_user = anon_user or AnonymousUserMixin
         self.cache = cache or FakeCache()
 
@@ -145,6 +157,10 @@ class SQLAlchemyBackend(BaseBackend):
         uid = first([user_id, self.user_id, blueprint.config.get("user_id")])
         u = first(_get_real_user(ref, self.anon_user)
                   for ref in (user, self.user, blueprint.config.get("user")))
+
+        if self.user_required and not u and not uid:
+            raise ValueError("Cannot get OAuth token without an associated user")
+
         # check for user ID
         if hasattr(self.model, "user_id") and uid:
             query = query.filter_by(user_id=uid)
@@ -166,6 +182,13 @@ class SQLAlchemyBackend(BaseBackend):
         return token
 
     def set(self, blueprint, token, user=None, user_id=None):
+        uid = first([user_id, self.user_id, blueprint.config.get("user_id")])
+        u = first(_get_real_user(ref, self.anon_user)
+                      for ref in (user, self.user, blueprint.config.get("user")))
+
+        if self.user_required and not u and not uid:
+            raise ValueError("Cannot set OAuth token without an associated user")
+
         # if there was an existing model, delete it
         existing_query = (
             self.session.query(self.model)
@@ -173,17 +196,12 @@ class SQLAlchemyBackend(BaseBackend):
         )
         # check for user ID
         has_user_id = hasattr(self.model, "user_id")
-        if has_user_id:
-            uid = first([user_id, self.user_id, blueprint.config.get("user_id")])
-            if uid:
-                existing_query = existing_query.filter_by(user_id=uid)
+        if has_user_id and uid:
+            existing_query = existing_query.filter_by(user_id=uid)
         # check for user (relationship property)
         has_user = hasattr(self.model, "user")
-        if has_user:
-            u = first(_get_real_user(ref, self.anon_user)
-                      for ref in (user, self.user, blueprint.config.get("user")))
-            if u:
-                existing_query = existing_query.filter_by(user=u)
+        if has_user and u:
+            existing_query = existing_query.filter_by(user=u)
         # queue up delete query -- won't be run until commit()
         existing_query.delete()
         # create a new model for this token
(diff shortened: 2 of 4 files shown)


NathanWailes:
Wow, that's great! That's exactly what I had in mind, except much-more-thoroughly done.

The one comment I would make is on `test_sqla_flask_login_no_user_required()`--and this is obviously assuming I'm reading it correctly--and it's that it doesn't seem to totally mirror the situation I was in because it looks like the test doesn't create a record in the OAuth table that has no `user_id`. That's what was happening to me: a record was being created in the OAuth table with no `user_id`, and then any logged-out user was being connected with that record.

NathanWailes:
Actually, now I'm wondering if raising an exception on the `get()` function is a good idea, because in my current implementation of Flask-Dance, the OAuth entry is created before the new `User` record is created. I don't remember the order in which these different functions are executed, but I feel like that `get()` function might excecute when the `current_user` is still `AnonymousUserMixin` (i.e. before the new `User` record is created and the user is logged in)?

singingwolfboy:
> That's what was happening to me: a record was being created in the OAuth table with no `user_id`, and then any logged-out user was being connected with that record.

I think that's a perfectly reasonable behavior, actually. If you have a User model, and you have an OAuth model with a foreign key to the User model, then doesn't it make sense that an OAuth entry in the database with `user_id` set to None should correspond to an anonymous user?

Lets say you want to build a website that is connected to Twitter, that works in the following way:

* It is connected to one particular account on Twitter. Anyone can visit your website, type in 140 characters, and post a tweet to that account without logging in.
* Your website also has optional user accounts. A user can connect their _personal_ Twitter account to your website. If a user is logged in to your website, she can tweet to her _personal_ Twitter account through the website, instead of the shared anonymous account.

This is perhaps a strange use-case, but a perfectly reasonable one in my opinion, and I'd like this library to be able to support that.

Particularly with the change in this pull request, that will make it more difficult to _accidentally_ create an OAuth entry with `user_id` set to None. If such an OAuth entry exists, then we can assume that it was created intentionally, to indicate that it should be used for anonymous users.

Does that seem reasonable? Do you think we can clarify this distinction with better documentation?

NathanWailes:
There's no way I understand this stuff as well as you do, so take all my comments below with a grain of salt:

---

> Lets say you want to build a website that is connected to Twitter, that works in the following way:

Off the top of my head, a way to implement it when a `user_id` is mandatory is to have any shared accounts as special User records. Because otherwise, how would you implement it when the number of 'public' / 'anonymous' Twitter accounts is greater than 1? The issue I'm discussing here is the ideal behavior of the `get()` function that gets called when you have something like `if twitter.authorized:` in your code, and AFAIK there's no way to specify which provider ID you want to get if the user is Anonymous, so it seems to me that the scenario you described would not be able to handle an Anonymous user choosing from more than one shared/public account. Disclaimer: I could just be misunderstanding / not thinking it through enough.

---

> Particularly with the change in this pull request, that will make it more difficult to accidentally create an OAuth entry with `user_id` set to None.

What's your take on the potential issue about `set()` that I raised in my previous comment? What you're describing here seems to depend on whether `set()` can be changed to force a `user_id` to be present.

singingwolfboy:
> There's no way I understand this stuff as well as you do, so take all my comments below with a grain of salt

The fact that you're new to this, and asking all these questions, is actually _incredibly helpful!_ Since I wrote this code, I'm _too_ familiar with it -- it's hard for me to see which things are confusing to other people. I want to make this project better, which means finding the confusing parts and making them less confusing. That's where you come in. The more questions you ask, and the more you tell me what is difficult and confusing about this library, the more I can address those difficulties and make it easier for _everyone_ to use Flask-Dance!

Also, as your understanding evolves, I'd love if you could submit more pull requests to the documentation. Write down the things you wish you had known when you started working with Flask-Dance, because those are the things that should go in the docs!

>  How would you implement it when the number of 'public' / 'anonymous' Twitter accounts is greater than 1?

You can't -- but that has nothing to do with anonymous users. The SQLAlchemy backend for Flask-Dance enforces a restriction of "not more than one OAuth entry for every user-provider pair". One user can only have one linked Twitter account. (A user can have more than one OAuth entry, as long as they're all different providers: one Twitter account _and_ one Facebook account is OK, but two Twitter accounts is not.) This is enforced via the [`one()`](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.one) method on the query. The anonymous user is just another user -- a user whose `user_id` happens to be None. You can't implement multiple Twitter accounts for the anonymous user, because you can't implement multiple Twitter accounts for _any_ user.

If you're writing an application that does need to support multiple Twitter accounts per user, you'll have to write a different Flask-Dance backend to support that. It's not even clear that the API that Flask-Dance exposes can support that, since as you point out, there's no good way of indicating _which_ Twitter account you want to retrieve for the user.

> It looks like the `set()` function is called when the provider redirects the user to the authorized endpoint.

Yep, if you allow Flask-Dance to handle saving the OAuth token for you, then it will call the `set()` function -- and if you don't yet have a User object, you have a problem. That's why you need to handle this yourself in the `oauth_authorized` signal handler, and return False so that Flask-Dance doesn't try to duplicate your work. As long as you create the User object and the OAuth object together in the same database transaction, and associate them with each other in that same transaction, you shouldn't have a problem.

> Also, with what you added to the `get()` function in this PR, wouldn't that make the scenario you described not work? Like, the Anonymous user would have an exception raised for them?

You would have to pass `require_user=False` to the SQLAlchemy backend, since you're allowing the anonymous user to get and set OAuth entries. As long as you do that, it won't raise an exception.

NathanWailes:
> It's not even clear that the API that Flask-Dance exposes can support that, since as you point out, there's no good way of indicating which Twitter account you want to retrieve for the user.

But what about the suggestion I gave? It seems to me that an alternative way to implement that functionality would be to have any anonymous user who wants to post from any of several 'public' Twitter accounts go to a page on your website that has a form, where the user would simply type the desired message into a box, select a drop-down of which of the different public Twitter accounts they want to post from, and the page would sent a POST request to the server with the desired Twitter account and message, and the server would then look up the OAuth credentials for the selected public User record and send a request to Twitter using those credentials.

---

> That's why you need to handle this yourself in the oauth_authorized signal handler

Ok I get it, thanks! I would maybe recommend making it clear that implementing that handler is now(?) mandatory if you're linking OAuth records to users. I'll take a look around the docs and see if I can find a good place for a warning like that.

---

> You would have to pass require_user=False to the SQLAlchemy backend, since you're allowing the anonymous user to get and set OAuth entries. As long as you do that, it won't raise an exception.

Got it, thanks! I forgot about that new parameter.

---

Before I forget: one thing that I remember tripped me up was the multiple contexts of the word "session", because I was thinking that "session" only referred to a user's session on my website, but then later I realized that the word session *also* refers to the session that exists between my website and the provider. I'm not really sure the best place to put that clarification, but I'll look around for a place in the docs to put it. I think a good way to clarify it might be to describe the sessions as "the user-consumer session" and "the consumer-provider session".

NathanWailes:
Alright I just spent an hour putting together a PR with some ideas, treat it like a first draft.

Also, I'm thinking it might be helpful to have quickstarts for the case where the developer needs to link oauth records to User accounts. If I have some free time I may take a stab at it, but I figured I'd let you know about it now.

elyselam:
Where exactly am I supposed to add "require_user=False" and how??

NathanWailes:
@elyselam If you look at the README, you should see the code below:

```
from flask_sqlalchemy import SQLAlchemy
from flask_dance.consumer.backend.sqla import OAuthConsumerMixin, SQLAlchemyBackend

db = SQLAlchemy()

class User(db.Model):
    id = db.Column(db.Integer, primary_key=True)
    # ... other columns as needed

class OAuth(OAuthConsumerMixin, db.Model):
    user_id = db.Column(db.Integer, db.ForeignKey(User.id))
    user = db.relationship(User)

# get_current_user() is a function that returns the current logged in user
blueprint.backend = SQLAlchemyBackend(OAuth, db.session, user=get_current_user)
```

Note the `db = SQLAlchemy()` line. That's what we're going to change.

The argument isn't actually called `require_user` now, he changed it to be `user_required`.

Unless I'm mistaken, the way you'd use it is to change the `db = SQLAlchemy()` line above to `db = SQLAlchemy(user_required=True)`.

You can see the code that refers to `user_required` [here](https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/consumer/backend/sqla.py).

singingwolfboy:
You're close, but that's not actually correct, @NathanWailes. You can see from the import statements that the `SQLAlchemy` class is imported from `flask_sqlalchemy`, not from this project -- so there's no way that this project could affect that class.

Instead, you should pass the `user_required` argument to the `SQLAlchemyBackend` class, which is imported from `flask_dance.consumer.backend.sqla`. Your code could look something like this:

```python
blueprint.backend = SQLAlchemyBackend(OAuth, db.session, user=get_current_user, user_required=False)
```

Note that by passing this argument, you are indicating that it's OK for OAuth tokens to be saved in the database without a user to go with that token. These tokens may never be used, and will just end up wasting space in the database.

PR #93, opened 2017-11-05
Documentation

NathanWailes:
Just some more ideas, obviously feel free to keep whatever's helpful and discard the rest.

- Just as a heads up, I'm a little nervous that my suggested session-related comments are not accurate.
- Also, I removed imports that PyCharm said were unused, so that's how I did that. It might not be right.


singingwolfboy:
Thanks!

PR #103, opened 2018-01-26
Added support for custom definition of url_for

giocalitri:
This is mostly a proposal to add a custom definition of `url_for` inside the `OAuth1ConsumerBlueprint` and `OAuth2ConsumerBlueprint`.

The reason for a custom `url_for` is that the flask `url_for` with the parameter `_external=True` does not behave well in a dockerized environment.

If this idea seems good, I am open to complete the PR and change also the other providers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - 2.7
   - 3.4
   - 3.5
+  - 3.6
 addons:
   postgresql: "9.4"
 env:
--- a/flask_dance/consumer/oauth1.py
+++ b/flask_dance/consumer/oauth1.py
@@ -39,6 +39,7 @@ class OAuth1ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             redirect_to=None,
             session_class=None,
             backend=None,
+            _url_for=None,
 
             **kwargs):
         """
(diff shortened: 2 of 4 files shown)


singingwolfboy:
> The reason for a custom `url_for` is that the flask `url_for` with the parameter `_external=True` does not behave well in a dockerized environment.

Can you elaborate on this? What problem are you seeing in this case?

giocalitri:
My flask app runs inside its own container on port 80 that is exposed outside the container on port 9000 that is exposed to the outside world on port 80 using nginx.

The only way I have found so far to create real external URL for my app is to set the external base url in a settings variable and then use a function like this:
```
def external_url_for(endpoint, **values):
    """
    Generates a URL to the given endpoint.
    This is needed because we cannot generate external URLs directly with url_for
    """
    # _external will be ignored
    if '_external' in values:
        values.pop('_external')
    return urljoin(settings.EXTERNAL_BASE_URL, url_for(endpoint, **values))
```

wgwz:
Does the flask `SERVER_NAME` parameter not help here?

giocalitri:
Setting `SERVER_NAME` is quite tricky in my environment.

inside docker my flask app is reachable as `http://api`, but this endpoint is exposed through an `nginx` container as `http://localhost:9000` so I can reach it with my browser.

Now if I set `SERVER_NAME` to `http://localhost:9000` (that is the "public" URL), the flask app starts to responds with 404s to any request.

I have no idea why the `SERVER_NAME` is messing with the rest of the configuration or how this could be solved.

singingwolfboy:
@giocalitri, I'm still not sure that I understand your problem. Can you make a reproducible test case, showing what you're trying to achieve and how it's failing? In this case, it sounds like that will be a minimal application using docker-compose.

A test case will help me understand the problem and suggest solutions. I doubt that replacing `url_for` is the best solution to your problem.

giocalitri:
Ok: I think I found the source of my problem and you guys were right: `SERVER_NAME` is what I should have looked at since the beginning.

PR #107, opened 2018-02-10
Fetch access_token from backend

### LINKED ISSUE #104: The github.access_token return None even if the access token is available
A demonstration is in this app:
https://github.com/giocalitri/flask-dance-github/blob/gdm_test/github.py

After getting the token with the first endpoint, try to GET the second and you will have something like:

```
{
  "access_token": null, 
  "token": {
    "access_token": "XXXXXXXXXXXXXXXXXXXXXXXXXX", 
    "scope": [
      ""
    ], 
    "token_type": "bearer"
  }
}
```

singingwolfboy:
Fixes #104. While investing that issue, I discovered that accessing the `access_token` property on an instance of the `OAuth2Session` class would only [attempt to locate an OAuth token on the client of the instance itself](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L116-L118), rather than querying the token backend.

Now, the `OAuth2Session` class defines its own `access_token` property that is aware of the token backend, and will use that backend instead of the client on the instance.

@giocalitri, can you test this out and let me know if it fixes your problem?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,9 @@ Change Log
 
 unreleased
 ----------
-Nothing yet
+* Accessing the ``access_token`` property on an instance of the
+  ``OAuth2Session`` class will now query the token backend, instead of
+  checking the client on the instance.
 
 0.13.0 (2017-11-12)
 -------------------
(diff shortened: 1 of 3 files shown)


giocalitri:
Yes, this fixes the issue.

PR #110, opened 2018-02-14
Bug: Fixed a small typo

nabinkhadka:
Fixes the following exception

`TypeError: 'name' is an invalid keyword argument for User`

### DIFF
--- a/docs/multi-user.rst
+++ b/docs/multi-user.rst
@@ -191,7 +191,7 @@ behavior in order to create user accounts automatically:
                 # Remember that `email` can be None, if the user declines
                 # to publish their email address on GitHub!
                 email=github_info["email"],
-                name=github_info["name"],
+                username=github_info["name"],
             )
             # Associate the new local user account with the OAuth token
             oauth.user = user


singingwolfboy:
Thanks for sending a pull request -- I really appreciate it! I'd love for more people to contribute to the documentation. :smile:

However, I don't think this change makes sense. The `User` model is defined by the developer writing the application, not by Flask-Dance. That model can have whatever fields the developer wants to define, including a "name" field for the user's name. That's what this example code is populating -- it get's the user's name from GitHub, and adds that to the "name" field in the local database. It's not about a username at all.

Does that make sense? Can you tell me why you thought this was supposed to be a username field?

nabinkhadka:
@singingwolfboy This is because of this one https://flask-dance.readthedocs.io/en/latest/multi-user.html

singingwolfboy:
@nabinkhadka, you just pointed to the rendered HTML version of the file you're modifying. I'm still not sure why you thought this was supposed to be a username field...

nabinkhadka:
@singingwolfboy Because I had followed the whole tutorial and I was previously reading [this](https://flask-dance.readthedocs.io/en/latest/quickstarts/sqla-multiuser.html) as well. The link has "username"

PR #112, opened 2018-02-26
modify seconds error for delta

ssoonan:
if _token["expires_in"] type isnt int, it throws an TypeError.

### DIFF
--- a/flask_dance/consumer/base.py
+++ b/flask_dance/consumer/base.py
@@ -108,7 +108,7 @@ class BaseOAuthConsumerBlueprint(six.with_metaclass(ABCMeta, flask.Blueprint)):
         if _token and _token.get("expires_in"):
             # Set the `expires_at` value, overwriting any value
             # that may already be there.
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Hi @ssoonan, can you tell me which provider is returning an `expires_in` value that isn't an int? That seems like a violation of the OAuth spec, and it's not something that this library should have to work around.

ssoonan:
thanks for your reply;

that provider is a Naver, which is Korea No.1 IT company.

Also, I didn't know that OAuth spec.

Thank you for your consideration.

2018-03-05 17:26 GMT+09:00 David Baumgold <notifications@github.com>:

singingwolfboy:
Here's where the spec defines this value: https://tools.ietf.org/html/rfc6749#appendix-A.14

> expires-in = 1*DIGIT

I think I'm going to close this pull request. As I said, if Naver is not compliant to the OAuth spec, then they need to fix that.

Issue #115, opened 2018-03-05
GitLab OAuth providor, client MissingTokenError before requesting token 


jgeorgeson:
I'm writing Flask app to drive some GitLab workflows that aren't supported by their API. I don't really need to have a user database/profile in my app. My app just needs to fetch the list of GitLab groups that the user has access to, they select a group, and then my app will make API calls and splinter/selenium calls to perform the workflow. I've configured my application in GitLab per this doc

https://docs.gitlab.com/ee/integration/oauth_provider.html#oauth-applications-in-the-admin-area

I've setup a OAuth2ConsumerBlueprint object with the client ID/secret and URLs for authorization and tokens, according to this doc

https://docs.gitlab.com/ee/api/oauth2.html

When I click on the `<a href={{url_for("oauth-gitlab.login")}}>Login with GitLab</a>` link in my app, it takes me to my GitLab instance to login and asks me to approve, then it redirects me back to my app and that throws `oauthlib.oauth2.rfc6749.errors.MissingTokenError`. 

```
File "/home/jgeorgeson/Code/Dev-Engineering/gitlab-import-wrapper/import-wrapper/lib/python3.6/site-packages/flask_dance/consumer/oauth2.py", line 234, in authorized
**self.token_url_params
File "/home/jgeorgeson/Code/Dev-Engineering/gitlab-import-wrapper/import-wrapper/lib/python3.6/site-packages/requests_oauthlib/oauth2_session.py", line 244, in fetch_token
self._client.parse_request_body_response(r.text, scope=self.scope)
File "/home/jgeorgeson/Code/Dev-Engineering/gitlab-import-wrapper/import-wrapper/lib/python3.6/site-packages/oauthlib/oauth2/rfc6749/clients/base.py", line 408, in parse_request_body_response
self.token = parse_token_response(body, scope=scope)
File "/home/jgeorgeson/Code/Dev-Engineering/gitlab-import-wrapper/import-wrapper/lib/python3.6/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 379, in parse_token_response
validate_token_parameters(params)
File "/home/jgeorgeson/Code/Dev-Engineering/gitlab-import-wrapper/import-wrapper/lib/python3.6/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 389, in validate_token_parameters
raise MissingTokenError(description="Missing access token parameter.")
```

[...]

PR #116, opened 2018-03-10
Gitlab provider

jgeorgeson:
In reference to #115 

Contrib requirements:
* [x] Provider file [`contrib/gitlab.py`](contrib/gitlab.py)
* [x] Declares `__maintainer__`
* [x] Factory function `make_gitlab_provider()` returning OAuth2ConsumerBlueprint
* [x] Provides local proxy object `gitlab`
* [x] Tests defined in [`tests/contrib/test_gitlab.py`](tests/contrib/gitlab.py), all tests pass
* [x] Updates CHANGELOG, README, and docs/providers


singingwolfboy:
This looks great! However, GitLab is open source, and can be deployed at many different domains. Can you add an argument to the make_gitlab_blueprint function for the domain to use, and make the argument default to “gitlab.com”? That way, this configuration is more useful.

jgeorgeson:
I actually meant to do that in the first place, can't believe I forgot it.

jgeorgeson:
Voilà.

GitHub also offers an self-hosted option, https://enterprise.github.com/home.

flask_dance/contrib/gitlab.py
```diff
@@ -0,0 +1,74 @@
...
+    from flask import _request_ctx_stack as stack
+
+
+__maintainer__ = "Justin Georgeson <jgeorgeson@lopht.net>"
+
+
+def make_gitlab_blueprint(
+        client_id=None, client_secret=None, scope=None, redirect_url=None,
+        redirect_to=None, login_url=None, authorized_url=None,
+        session_class=None, backend=None, hostname="gitlab.com"):
```
singingwolfboy: GitLab's biggest value proposition over GitHub is that you can run it on your own server, so I expect that this `hostname` parameter will be widely used. What do you think about moving it to the front of the arguments list, so that people could call it as a positional argument? For example:

```python
gitlab_bp = make_gitlab_blueprint("my-gitlab-installation.com")
```

flask_dance/contrib/gitlab.py
```diff
@@ -0,0 +1,74 @@
...
+            specify the hostname, default is ``gitlab.com``
+
+    :rtype: :class:`~flask_dance.consumer.OAuth2ConsumerBlueprint`
+    :returns: A :ref:`blueprint <flask:blueprints>` to attach to your Flask app.
+    """
+    gitlab_bp = OAuth2ConsumerBlueprint("gitlab", __name__,
+        client_id=client_id,
+        client_secret=client_secret,
+        scope=scope,
+        base_url="https://"+hostname+"/api/v4/",
```
singingwolfboy: I strongly prefer using [string formatting](https://docs.python.org/3/library/string.html#format-string-syntax) rather than concatenation like this. String formatting always coerces variables to strings before putting them together, while concatenation with the plus sign can throw exceptions if the variable is not already a string. You could do something like this:

```python
base_url = "https://{hostname}/api/v4".format(hostname=hostname)
```

However, in this specific case, you have an even better tool available. The [urlobject](https://urlobject.readthedocs.io/en/latest/quickstart.html) library is an elegant way of solving _precisely_ this problem, and Flask-Dance already depends on it. You could do something like this:

```python
base_url = URLObject("https://gitlab.com/api/v4/").with_hostname(hostname)
```

What do you think of these solutions?

tests/contrib/test_gitlab.py
```diff
@@ -0,0 +1,93 @@
...
+        redirect_to="index"
+    )
+    assert isinstance(glbp1, OAuth2ConsumerBlueprint)
+    assert glbp1.session.scope == "read_user"
+    assert glbp1.session.base_url == "https://gitlab.com/api/v4/"
+    assert glbp1.session.client_id == "foo"
+    assert glbp1.client_secret == "bar"
+    assert glbp1.authorization_url == "https://gitlab.com/oauth/authorize"
+    assert glbp1.token_url == "https://gitlab.com/oauth/token"
+    # Test with custom hostname
```
singingwolfboy: Could you divide this into two separate test functions, please?

flask_dance/contrib/gitlab.py
singingwolfboy: > GitHub also offers an self-hosted option, https://enterprise.github.com/home.

Hmm, that's true. (Didn't see that comment while I was reviewing this PR.) And now that I think about it, if we move this `hostname` parameter to the front, then that creates a different call signature compared to the `make_github_blueprint` factory -- and while we _could_ change the call signature of that factory as well, that would be a backwards-incompatible change, which I'd like to avoid.

What do you think? I see pros and cons to either approach.

jgeorgeson: Putting it at the front was my initial instinct but I thought uniformity of the signature compared to the other providers, especially with respect to GitHub provider, was reason enough not to. But I'm fine with moving it to the front if you still prefer.

flask_dance/contrib/gitlab.py
jgeorgeson: I prefer using the `format()` function over adding an additional dependency, will make the change.

tests/contrib/test_gitlab.py
jgeorgeson: will do.

singingwolfboy:
Thanks! I'll cut a new release in the next few days, and push it to PyPI. :tada:

singingwolfboy:
[Flask-Dance 0.14.0 is out!](https://pypi.python.org/pypi/Flask-Dance/0.14.0) 🎆

PR #127, opened 2018-05-16
Added try except to parse_authorization_response in oauth1

chuyachia:
When entered Twitter Oauth page and clicked"Cancel" instead of "Authorize app", Twitter propose a button to return to the original app that redirects to "/twitter/authorized?denied=xxxx".
This calls the authorized method in oauth1.py. However, the fact that the redirect url only contains a denied key and does not have an oauth_token key causes self.session.parse_authorization_response(request.url) to crash.
Adding try except around self.session.parse_authorization_response(request.url) avoids this crash.

### DIFF
--- a/flask_dance/consumer/oauth1.py
+++ b/flask_dance/consumer/oauth1.py
@@ -202,8 +202,9 @@ class OAuth1ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             self.session.parse_authorization_response(request.url)
         except TokenMissing as err:
             message = err.args[0]
+            response = getattr(err, "response", None)
             log.warning("OAuth 1 access token error: %s", message)
-            oauth_error.send(self, message=message)
+            oauth_error.send(self, message=message, response=response)
             return redirect(next_url)
 
         try:
--- a/flask_dance/consumer/oauth1.py
+++ b/flask_dance/consumer/oauth1.py
@@ -4,7 +4,7 @@ import logging
 from lazy import lazy
 from flask import request, url_for, redirect, current_app
 from werkzeug.wrappers import Response
-from requests_oauthlib.oauth1_session import TokenRequestDenied
+from requests_oauthlib.oauth1_session import TokenRequestDenied, TokenMissing
 from oauthlib.oauth1 import SIGNATURE_HMAC, SIGNATURE_TYPE_AUTH_HEADER
 from oauthlib.common import to_unicode
 from .base import BaseOAuthConsumerBlueprint, oauth_authorized, oauth_error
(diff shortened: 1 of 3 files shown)


flask_dance/consumer/oauth1.py
```diff
@@ -197,7 +197,15 @@ def authorized(self):
             next_url = url_for(self.redirect_to)
         else:
             next_url = "/"
-        self.session.parse_authorization_response(request.url)
+            
+        try:
+            self.session.parse_authorization_response(request.url)
+        except Exception as err:
```
singingwolfboy: It seems like this should raise a `TokenMissing` exception: https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth1_session.py#L352

Can you change this generic `Exception` type to that more specific `TokenMissing` exception?

flask_dance/consumer/oauth1.py
```diff
@@ -197,7 +197,15 @@ def authorized(self):
...
+            message = err.args[0]
+            response = getattr(err, "response", None)
```
singingwolfboy: This exception will never have a `response` property, so you can remove this.

chuyachia:
It seems like that the automated test codes failed CI in python 2.7 because of unicode strings. I will continue to work on it. Other than this problem, does the automated test look good to you?

singingwolfboy:
Yep, looks good! I suggest that you change the log line to `log.warning("OAuth 1 access token error: %s", json.dumps(message))` so that the strings are encoded without the `u` prefixes.

flask_dance/consumer/oauth1.py
```diff
@@ -209,7 +216,7 @@ def authorized(self):
             message = err.args[0]
             response = getattr(err, "response", None)
             log.warning("OAuth 1 access token error: %s", message)
-            oauth_error.send(self, message=message, response=response)
+            oauth_error.send(self, message=message)
```
singingwolfboy: Please revert this change. _This_ error _will_ have a `response` property, and it should be passed along.

chuyachia:
Applying `json.dumps` to `message` does not seem to remove the u prefixes as `message` is a string formed by string formatter that would already incorporate the u prefixes under Python 2
To remove the u prefixes, `json.dumps` would have to be applied here in the requests-oauthlib module https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth1_session.py#L353 with something like `'Response does not contain a token: {resp}'.format(resp=json.dumps(token))` 

One workaround without changing the requests-oauthlib module would be to subsitute the `{'denied':'xxx'}` part in the `message` string in oauth1's `authorized ` method but I think we could just ignore the `{'denied':'xxx'}` part when we check `message` and check it separately in `response` in the automated test.

PR #139, opened 2018-06-22
Updated docs. Currently unclear as to whether the `scopes` value shou…

JellyWX:
…ld be a string or a list of strings


singingwolfboy:
Thanks!

PR #141, opened 2018-07-03
docs: Add documentation on testing strategies

daenney:
It took me a while to figure out how to be able to test a Flask app that had some or all of its routes behind a login decorator mandating oauth2.

This is a write-up of what one could do. I'm hoping to get this accepted into flask-dance as I believe this would prove useful to others.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,0 +1,155 @@
+Testing
+=======
+
+When building Flask apps that have integrated Flask-Dance you'll eventually run
+into the need to test your app. But if routes are guarded by the requirement of
+having a valid OAuth2 session, i.e getting past the ``.authorized()`` call on
+a provider, how do you test this?
+
+One possible way of solving this is by injecting the necessary information into
+the Flask session. The following examples assume you're using `py.test`_ and the
+Google provider, but the same trick applies to any of the supported providers.
+
+Creating a session
+------------------
+
+In your ``conftest.py`` create two fixtures, like this:
+
+.. code-block:: python
+
+    import time
+
+    import pytest
+
+    from your_app import create_app
+    from your_app.settings import Testing
+
+    fake_time = time.time()
+
+
+    @pytest.fixture
+    def app():
+        """Returns an app fixture with the testing configuration."""
+        app = create_app(config_object=Testing)
+        yield app
+
+
+    @pytest.fixture
+    def loggedin_app(app):
+        """Creates a logged-in test client instance."""
+        with app.test_client() as client:
+            with client.session_transaction() as sess:
+                sess['google_oauth_token'] = {
+                    'access_token': 'this is totally fake',
+                    'id_token': 'this is not a real token',
+                    'token_type': 'Bearer',
+                    'expires_in': '3600',
+                    'expires_at': fake_time + 3600,
+                }
+            yield client
+    
+This will inject the necessary information in order to ensure that Flask-Dance,
+requests-oauthlib and oauthlib believe there is a valid session, and one that
+won't need to be refreshed for another hour (3600 seconds). If your tests run
+longer than that, you'll need to adjust that value.
+
+The ``fake_time`` is created so that you can always refer to the same point in
+time throughout tests, and in any other fixture you might create. Alternatively
+you can use something like `pytest-freezegun`_ and then call any of the ``time``
+and ``datetime`` functions as you normally would:
+
+.. code-block:: python
+
+    @pytest.fixture
+    @pytest.mark.freeze_time('2018-05-04')
+    def loggedin_app(app):
+        """Creates a logged-in test client instance."""
+        with app.test_client() as client:
+            with client.session_transaction() as sess:
+                sess['domain'] = 'example.com'
+                sess['google_oauth_token'] = {
+                    'access_token': 'this is totally fake',
+                    'id_token': 'this is not a real token',
+                    'token_type': 'Bearer',
+                    'expires_in': '3600',
+                    'expires_at': time.time() + 3600,
+                }
+            yield client
+
+Now that we have a logged-in client you can call any of the routes using the
+test client and check their responses.
+
+.. code-block:: python
+
+    def test_not_logged_in(app):
+        """Test that we redirect to Google to login."""
+        res = app.test_client().get('/')
+        assert ('redirected automatically to target URL: <a href="/login/'
+                'google">/login/google</a>').lower() in res.get_data(as_text=True).lower()
+        assert res.status_code == 302
+
+
+    def test_logged_in_index(loggedin_app):
+        """Tests getting the index route.
+
+        This will render the normal template as we have a valid oauth2 session.
+        """
+        res = loggedin_app.get('/')
+        assert res.content_type == 'text/html; charset=utf-8'
+        assert res.status_code == 200
+        assert 'something only shown when logged in' in res.get_data(as_text=True).lower()
+
+
+Calling authenticated APIs
+--------------------------
+
+Though we've managed to create a working session a problem now arises if you
+try to actually call an API, by using ``google.get('some url')`` for example.
+Your token will fail to validate and the request will be denied.
+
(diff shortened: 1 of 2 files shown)


singingwolfboy:
This looks great! However, you've made a few reStructuredText syntax mistakes -- it's a different syntax from Markdown. [Here's a reStructuredText primer, which explains the syntax.](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) In addition, there are a few places where it would be really handy to link to the project that you mention in the documentation. So far, I've tried to use named references, like this:

```rst
...assume you're using `py.test`_ ...

.. Then, at the bottom of the file:

.. _`py.test`: https://docs.pytest.org/
```

I'll go through the diff, and add comments to the specific lines that need some changes. But overall, this is a fantastic contribution, and I can't wait to get this merged! 😄

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
+Testing
+=======
+
+When building Flask apps that have integrated Flask-Dance you'll eventually run
+into the need to test your app. But if routes are guarded by the requirement of
+having a valid oauth2 session, i.e getting passed the `.authorized()` call on
```
singingwolfboy: "OAuth" should be capitalized. You're using "passed" instead of "past" -- the correct spelling should be "getting past". Also, fixed-width text requires two sets of backticks in rst: 

    ``.authorized()``

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+a provider, how do you test this?
+
+One possible way of solving this is by injecting the necessary information into
+the Flask session. The following examples assume you're using py.test and the
```
singingwolfboy: Please link to the py.test website.

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+Google provider, but the same trick applies to any of the supported providers.
+
+Creating a session
+------------------
+
+In your `conftest.py` create two fixtures, like this:
```
singingwolfboy: Fixed-width text requires two sets of backticks in rst:

    ``conftest.py``

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+                    'access_token': 'this is totally fake',
+                    'id_token': 'this is not a real token',
+                    'token_type': 'Bearer',
+                    'expires_in': '3600',
+                    'expires_at': fake_time + 3600,
+                }
+            yield client
+    
+This will inject the necessary information in order to ensure that Flask-Dance,
+requests-oauthlib and oauthlib2 believe there is a valid session, and one that
```
singingwolfboy: It's just `oauthlib`, not `oauthlib2`

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+won't need to be refreshed for another hour (3600 seconds). If your tests run
+longer than that, you'll need to adjust that value.
+
+The `fake_time` is created so that you can always refer to the same point in
```
singingwolfboy: Fixed-width text requires two sets of backticks in rst:

``fake_time``

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+time throughout tests, and in any other fixture you might create. Alternatively
+you can use something like pytest-freezegun and then call any of the `time` and
```
singingwolfboy: Please link to the pytest-freezegun website. Also, `time` and `datetime` need two sets of backticks.

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+
+
+Calling authenticated APIs
+--------------------------
+
+Though we've managed to create a working session a problem now arises if you
+try to actually call an API, by using `google.get('some url')` for example.
+Your token will fail to validate and the request will be denied.
+
+This can be handled by a Python library called responses, which lets us control
```
singingwolfboy: `responses` is a very useful library, and it's what I used to write tests for Flask-Dance itself. However, _after_ I wrote those tests, I found out about another library called [`betamax`](https://github.com/betamaxpy/betamax), which is written by one of the maintainers of Requests. It's a really nice library for intercepting HTTP requests and replaying them in tests, but it's more complicated to set up.

You might want to look into `betamax`, and see if it fits your usecase. If it does, and you're inspired to rewrite this example using `betamax` instead of `responses`, let me know! If not, or if you're happy with the documentation referring to `responses`, then that's fine -- I'm happy to get this pull request merged either way. (Just wanted to tell you about this cool library!)

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+the full HTTP request cycle. Note that this means we're essentially mocking the
+API we're calling, so your tests will continue passing even if the real API
```
singingwolfboy: You might want to use the `warning` directive here. For example:

```rst
.. warning::
    Note that this means we're essentially mocking the API we're calling,
    so your tests will continue passing even if the real API has changed behavior.
```
https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#colored-boxes-note-seealso-todo-and-warnings

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+has changed behaviour.
+
+Lets assume the index route calls out to the Google Plus API and displays some
```
singingwolfboy: Missing an apostrophe: "Let's". Also, could you link to the [Google Plus API](https://developers.google.com/+/web/api/rest/)?

docs/testing.rst
```diff
@@ -0,0 +1,140 @@
...
+
+
+    @responses.activate
+    def test_getting_profile(loggedin_app):
+        """Test displaying profile information."""
+        responses.add(
+            responses.GET,
+            'https://www.googleapis.com/plus/v1/people/me',
+            status=200,
+            json={'whatever this API would return'})
```
singingwolfboy: This is not valid JSON. I understand that this is just a placeholder, but could you use valid JSON, so that the test is at least theoretically runnable?

docs/testing.rst
daenney: I had the same experience, became aware of betamax after doing this. I'd lprefer to submit a PR updating or adding a betamax example later on, once I've actually had some time to use it.

daenney:
I think this fixes all of it. Let me know if there's anything else.

docs/testing.rst
singingwolfboy: Totally reasonable! Let me know if I can help or answer questions about Betamax and Flask-Dance.

docs/testing.rst
```diff
@@ -0,0 +1,155 @@
...
+            responses.GET,
+            'https://www.googleapis.com/plus/v1/people/me',
+            status=200,
+            json={
+              'kind': 'plus#person',
+              'id': '118051310819094153327',
+              'displayName': 'Chirag Shah',
+              'url': 'https://plus.google.com/118051310819094153327',
+              'image': {
+                'url': 'https://lh5.googleusercontent.com/-XnZDEoiF09Y/AAAAAAAAAAI/AAAAAAAAYCI/7fow4a2UTMU/photo.jpg'
```
singingwolfboy: I just want to confirm that you're OK with having this content available in the public documentation for Flask-Dance. Some people might consider a name and photograph to be private information. Is this OK with you?

daenney: Ah. This is a straight up copy from the Google Plus documentation: https://developers.google.com/+/web/api/rest/#data-formats. So from my p.o.v it's fine. I'm assuming it's some test data they purposefully created to have a more usable example.

singingwolfboy: Ah, if Google has decided it's OK, then I'm fine with that. 👍

daenney:
I found a few other places with missing double backticks and a missing link. I think that's all of it now.

singingwolfboy:
Thanks! Here's the live version: http://flask-dance.readthedocs.io/en/latest/testing.html :tada:

daenney:
Awesome. Thanks for the quick feedback and turnaround time! It's great when maintainers are so responsive.

PR #144, opened 2018-07-12
Update to testing and Google quickstart

### LINKED ISSUE #143: refresh_token and online vs offline access
This is kind of a rehash of #137 and #111 and #102. I'm hoping to get some input here (and then maybe provide a documentaiton PR). This appears specific to the Google provider.

Accroding to Google's OAuth documentation you should request an access token of type offline if your applicaiton needs to do actions on behalf of the user while the user is offline. Say for example some kind of scheduled job that runs at some point in time, like backups.

Based on that documentation, if you're just writing an app where you're acting on behalf of the user while she is online, i.e sitting at their browser, it wouldn't occur to you to request an offline token. However, if the user opens up your app, does a thing, goes away for an hour and then tries to use it again you'll hit the `Missing required parameter: refresh_token` because at this point the token has expired and you need to grab a new one.

So far what I've seen is that people just change the blueprint call to fetch an offline token instead, but that seems incorrect to me, since you're not really doing anything offline. This also doesn't really work for session storage b/c as far as I can tell the refresh token is never stored in the user session, it should be saved in a server side backend. The user is still there, it's just that they had other stuff to do and now the token has expired.

From my point of view the solution in #137 is correct, add an error handler to catch the `InvalidClientIdError` and redirect the user. This'll retrigger the OAuth flow and when users have already given consent won't actually do much of anything, essentially logging them back in with a fresh and valid token. No need to set `offline=true` in such cases.

The question here is, does this sound reasonable, and should we perhaps add documentation to the Google provider to reflect this?

daenney:
Fixes #143

### DIFF
--- a/docs/quickstarts/google.rst
+++ b/docs/quickstarts/google.rst
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
 :meth:`~requests.Session.get` and :meth:`~requests.Session.post`,
 to make HTTP requests. If you only specify the path component of the URL,
 the domain will default to ``https://www.googleapis.com``.
+
+Online vs offline applications
+------------------------------
+
+Google distinguishes between online and offline applications. Online applications
+are applications that only act when the user is at their keyboard, say a web
+application that only calls Google API's as the user is interacting with your
+application. Offline applications are applications that can act on behalf of the
+user while they're not at their keyboard, for example something that runs backups
+on behalf of the user at a specific point in time.
+
+The blueprint is configured by default to get an online token. This means you'll get
+a token that's valid for about an hour and after that you'll need to get a new token,
+by putting the user through the OAuth flow again. If instead you request an offline
+token you'll also be given a refresh token that your application can use to request
+a new, short lived, token without the user needing to do anything.
+
+.. warning::
+
+   The refresh token is only returned the first time you put a user through the OAuth
+   flow. This means you need to store the refresh token in a persistent fashion in
+   order to ensure your application can renew the access token. It's therefore not
+   advisable to store this using Flask's regular session, which uses a cookie. Though
+   the session can be marked as persistent that won't save you if the user decides
+   to clear their browser cache for example.
+
+It's tempting to just request an offline token but if your application doesn't act
+on behalf of the user while they're not at their keyboard this is in bad form. Instead,
+you can install an error handler in your Flask app that will automatically retrigger the
+OAuth flow when the token has expired.
+
+.. code-block:: python
+
+    from oauthlib.oauth2.rfc6749.errors import InvalidClientIdError
+
+    @app.errorhandler(InvalidClientIdError)
+    def token_expired(_):
+        """Get a fresh access token by triggering the OAuth flow.
+
+        Since the user has already given consent this won't cause the user to
+        have to interact with anything. In most cases it'll flash by without
+        the user noticing anything.
+        """
+        session.pop('google_oauth_token')
+        return redirect(url_for('index'))
+
+You'll need to adjust the ``redirect`` call to a URL that'll trigger the OAuth flow. You
+can also manually handle the error at the callsite instead, anywhere you do a ``google.get()``
+call for example, by catching the exception. That would allow you to customize the behaviour.
+In all cases you'll need to remove the ``google_oauth_token`` from the session and redirect
+to something that'll retrigger the OAuth flow.
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -60,24 +60,32 @@ and ``datetime`` functions as you normally would:
 
 .. code-block:: python
 
+    from freezegun import freeze_time
+
     @pytest.fixture
-    @pytest.mark.freeze_time('2018-05-04')
     def loggedin_app(app):
         """Creates a logged-in test client instance."""
-        with app.test_client() as client:
-            with client.session_transaction() as sess:
-                sess['domain'] = 'example.com'
-                sess['google_oauth_token'] = {
-                    'access_token': 'this is totally fake',
-                    'id_token': 'this is not a real token',
-                    'token_type': 'Bearer',
-                    'expires_in': '3600',
-                    'expires_at': time.time() + 3600,
-                }
-            yield client
+        with freeze_time('2018-05-04', tz_offset=0)
+            with app.test_client() as client:
+                with client.session_transaction() as sess:
+                    sess['domain'] = 'example.com'
+                    sess['google_oauth_token'] = {
+                        'access_token': 'this is totally fake',
+                        'id_token': 'this is not a real token',
+                        'token_type': 'Bearer',
+                        'expires_in': '3600',
+                        'expires_at': time.time() + 3600,
+                    }
+                yield client
 
 Now that we have a logged-in client you can call any of the routes using the
 test client and check their responses.
+
+.. note::
+    Using ``freeze_time`` time is now permanently frozen in this fixture.
+    If you want to move time forwards or backwards you'll need to accept the
+    ``freezer`` fixture in your tests, and call ``freezer.move_to()`` passing
+    in an `ISO 8601`_ formatted datetime string to move time to that point in time.
 
 .. code-block:: python
 
@@ -153,3 +161,4 @@ documentation instead.
 .. _`pytest-freezegun`: https://github.com/ktosiek/pytest-freezegun
 .. _`responses`: https://github.com/getsentry/responses
 .. _`Google Plus API`: https://developers.google.com/+/web/api/rest/
+.. _`ISO 8601`: https://en.wikipedia.org/wiki/ISO_8601


docs/quickstarts/google.rst
```diff
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
...
+
+   The refresh token is only returned the first time you put a user through the OAuth
+   flow. This means you need to store the refresh token in a persistent fashion in
+   order to ensure your application can renew the access token. It's therefore not
+   advisable to store this using Flask's regular session, which uses a cookie. Though
+   the session can be marked as persistent that won't save you if the user decides
+   to clear their browser cache for example.
+
+It's tempting to just request an offline token but if your application doesn't act
+on behalf of the user while they're not at their keyboard this is in bad form. Instead,
```
singingwolfboy: "Bad form" is not a good reason to avoid offline tokens. Instead, it's better to talk about security. A core principal of security is to only provide the minimal set of permissions necessary to get the job done. If your application doesn't need offline access, then it shouldn't request it, because offline tokens are a greater target for attackers. If you only use online tokens, that makes it much harder for an attacker to get access to the user's account.

docs/quickstarts/google.rst
```diff
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
...
+by putting the user through the OAuth flow again. If instead you request an offline
+token you'll also be given a refresh token that your application can use to request
+a new, short lived, token without the user needing to do anything.
+
+.. warning::
+
+   The refresh token is only returned the first time you put a user through the OAuth
+   flow. This means you need to store the refresh token in a persistent fashion in
+   order to ensure your application can renew the access token. It's therefore not
+   advisable to store this using Flask's regular session, which uses a cookie. Though
```
singingwolfboy: This is ambiguous. Flask-Dance includes two different token storage backends: the session backend, and the SQLAlchemy backend. (And others could write more backends in the future!) This is only relevant if the application is using the session backend, and if Flask sessions are implemented using HTTP cookies.

docs/quickstarts/google.rst
```diff
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
...
+
+    @app.errorhandler(InvalidClientIdError)
+    def token_expired(_):
+        """Get a fresh access token by triggering the OAuth flow.
+
+        Since the user has already given consent this won't cause the user to
+        have to interact with anything. In most cases it'll flash by without
+        the user noticing anything.
+        """
+        session.pop('google_oauth_token')
```
singingwolfboy: Rather than referencing the string `google_oauth_token` (which is a private implementation detail), it's clearer to just tell the token storage backend that the token should be removed. To do this, delete the token from the blueprint, like this:

```python
del google_bp.token
```

Can you verify that it works correctly?

docs/quickstarts/google.rst
```diff
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
...
+
+    from oauthlib.oauth2.rfc6749.errors import InvalidClientIdError
+
+    @app.errorhandler(InvalidClientIdError)
+    def token_expired(_):
+        """Get a fresh access token by triggering the OAuth flow.
+
+        Since the user has already given consent this won't cause the user to
+        have to interact with anything. In most cases it'll flash by without
+        the user noticing anything.
```
singingwolfboy: The user will probably notice an unexpected redirect to the application homepage. :(

docs/quickstarts/google.rst
```diff
@@ -95,3 +95,54 @@ you can use all the normal ``requests`` methods, like
...
+        """
+        session.pop('google_oauth_token')
+        return redirect(url_for('index'))
+
+You'll need to adjust the ``redirect`` call to a URL that'll trigger the OAuth flow. You
```
singingwolfboy: You could redirect to `google.login`...

daenney:
Using `refresh_token` isn't possible when you have an online token, that's part of the problem discussed here and why the developer needs to write that extra code. Those code paths already trigger.

singingwolfboy:
I realize that we can't use `refresh_token` for the online token flow, because Google doesn't provide a `refresh_token`. But my point is, the flow you're proposing in this documentation is pretty similar conceptually: the application automatically fetches a new token by bumping the user through the OAuth dance again.

I'm wondering if it's possible to _write some new code_, either for Flask-Dance or for Requests-OAuthlib, that provides hooks to handle this "online auto-dance" scenario in a similar way from how Requests-OAuthlib provides hooks to handle the `refresh_token` scenario. Does that seem possible?

daenney:
Aah, now I see what you're getting at. Apologies.

I suppose it is possible. We could package up the errorhandler for starters so that it can just be installed, instead of people needing to copy that code?

When it comes to handling the full flow, I believe it's possible. The `TokenExpiredError` bubbles up when you use the `googe.get()` etc methods so in the case of the `Missing required parameter: refresh_token` error, we could redo the OAuth flow. We'd have to introspec the error though, since ClientIdError can be raised for more than one thing. It would be helpful if requests-oauthlib raised a different error though, so we don't have to introspect the msg itself to know if it's a refresh token related problem.

What I'm unsure of is where to handle this. From what I can see we'd have to catch the exception in Flask-Dance, b/c only at that point do we have access to Flask's session object, requests-oauthlib can't do that. I'm not sure where we'd do that though, I'm guessing https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/consumer/requests.py#L136?

docs/quickstarts/google.rst
daenney: I suppose that's true. You can avoid that by handling the error yourself and including a redirect parameter when retriggering the OAuth flow.

docs/quickstarts/google.rst
daenney: How would one get to the `google_bp` when using an application factory style setup? Something along the lines of `current_app.blueprint...`?

daenney:
There is possibly another option, by pulling the token expiry checking into Flask-Dance.

I'm not sure if this works, but what if in a `before_request` we'd check if the token hadn't expired yet? Should it have lapsed and no refresh token be available, we could then trigger the OAuth flow somehow. The one thing I'm kinda stuck on is how do we do this fully transparently for the user? I.e, they don't notice anything and we don't end up redirecting them away from what they're doing?

docs/quickstarts/google.rst
daenney: Ah, `del current_app.blueprints['google'].token` does the trick.

singingwolfboy:
> We could package up the errorhandler for starters so that it can just be installed, instead of people needing to copy that code?

That's definitely a good start!

> I'm not sure if this works, but what if in a `before_request` we'd check if the token hadn't expired yet? Should it have lapsed and no refresh token be available, we could then trigger the OAuth flow somehow.

Hmm. There's a lot that I like about this idea, but the one big problem is that it's too eager. If the view needs to make OAuth-based requests, it's great, but if the view _doesn't_ need to make OAuth-based requests, it's doing extra work for no reason. At worst, it could cause some sort of error if the provider is misconfigured, or cause the user to have to re-click the authorize button on the provider website, both of which are very disruptive to a user flow!

>  It would be helpful if requests-oauthlib raised a different error, so we don't have to introspect the msg itself to know if it's a refresh token related problem.

Fortunately, we can make changes to `requests-oauthlib` as well! You're right that `requests-oauthlib` can't handle this situation all by itself, because it doesn't have any knowledge of Flask (nor should it). But we can handle this in Flask-Dance, and in any situation where it would be convenient to modify requests-oauthlib so that Flask-Dance has the information it needs, we can totally do that. (I'm a maintainer for requests-oauthlib too, which makes this easier.)

> The one thing I'm kinda stuck on is how do we do this fully transparently for the user? I.e, they don't notice anything and we don't end up redirecting them away from what they're doing?

Well, conceptually we need to do the following things, in order:

* Run a normal, OAuth-authorized request to the provider
* If the OAuth token is no longer valid, save the URL of the current page somewhere we can find it again, and redirect the user to go through the OAuth dance again
* At the end of the dance, catch and save the new token (just like we already do)
* After the token is saved, check to see if we've saved the URL of a page where we tried to do an OAuth request, but failed due to an invalid token
* If so, redirect the user to _that_ page, instead of the normal post-authorization redirect

This does mean that if the page view has any logic before the OAuth-authorized request is run, that logic will be repeated when the user is redirected back to that page with a new OAuth token -- but I don't think there's any way around that.

Does that help you come up with some ideas for how to implement this? I'm just bouncing ideas here -- I'm hoping that we can figure this out together, since I don't have all the answers!

daenney:
> If the view needs to make OAuth-based requests, it's great, but if the view doesn't need to make OAuth-based requests, it's doing extra work for no reason. At worst, it could cause some sort of error if the provider is misconfigured, or cause the user to have to re-click the authorize button on the provider website, both of which are very disruptive to a user flow!

Mm, indeed. So the question then becomes, how do we know? Do we provide a decorator instead that marks that route as doing things that will require OAuth?

> Run a normal, OAuth-authorized request to the provider
> If the OAuth token is no longer valid, save the URL of the current page somewhere we can find it again, and redirect the user to go through the OAuth dance again

This seems pretty doable, at least for Google there's a `tokeninfo` endpoint we can safely hit.

I wonder what happens in the case of POST routes though. Say someone fills in a form, hits post and then the form data is used to build and make the request to an OAuth protected endpoint. We'd need to capcture more than just hte URL to not break things like this.

singingwolfboy:
> Do we provide a decorator instead that marks that route as doing things that will require OAuth?

That seems like a solution that's worse than the problem. :(

> at least for Google there's a `tokeninfo` endpoint we can safely hit.

I intended to determine if the OAuth token was valid by just checking if there was an OAuth-related error in the response. Making a request to a `tokeninfo` endpoint before _every_ OAuth-protected response seems like a big performance problem.

> I wonder what happens in the case of POST routes though.

Yeah, it may not be possible to make that a seamless flow. Frankly, if there's an OAuth token problem in a flow like that, trying to handle it automatically may just introduce more problems -- I can imagine that debugging a situation like that would be hell. GET requests are much easier: maybe we should just try to handle those, and document that developers will need to handle POST requests manually.

daenney:
> I intended to determine if the OAuth token was valid by just checking if there was an OAuth-related error in the response. Making a request to a tokeninfo endpoint before every OAuth-protected response seems like a big performance problem.

Mmm, but how do we do that in a before_request style solution? At htat point the request hasn't triggered yet, right? So how can we check there's an error?

singingwolfboy:
It doesn't have to be a `before_request` style solution. The `google` object is an instance of the `OAuth2Session` class, and we can override methods on that class however we want. We can make `requests-oauthlib` raise a specific exception if there is an OAuth-related error in the response, then make `OAuth2Session.get` catch that exception and perform other logic as necessary.

daenney:
Aha. While investigating this a bit something occured to me, we should never get the `Missing required parameter refresh_token` in the first place, because requests-oauthlib shouldn't be trying to refresh an online access token (since that can never ever happen). It seems that when `offline=False` we should unset the `auto_refresh_url`: https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/contrib/google.py#L69.

Without `auto_refresh_url` set the condition over here would never be `True`: https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L333. That would put us inside https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L353-L354 where we could raise its own error, something like `UnrefreshableTokenExpiredError` (I'm terrible at naming things)? That in turn we could then properly catch in the case of an `OAuth2Session.get` and handle, and in case of a `.post`... reraise?

singingwolfboy:
> It seems that when `offline=False` we should unset the `auto_refresh_url`

That seems reasonable to me! And I don't think we need to think of a new name and raise a new error: `TokenExpiredError` is accurate. The fact that the error is being raised from requests-oauthlib indicates that it can't be automatically refreshed, so the "unrefreshable" part is implied. :)

> That in turn we could then properly catch in the case of an `OAuth2Session.get` and handle, and in case of a `.post`... reraise?

That also seems reasonable. We actually don't need to put any special handling in `.post` at all -- no need to catch an exception if you simply intend to re-raise it. Maybe we can add something to the documentation to indicate that the developer should add code to handle this case, if desired -- sometimes the best solution is code, and sometimes the best solution is documentation. :smile:

Do you want to open a new pull request to make these changes? From what you've described, I don't think we actually need to make any changes to requests-oauthlib at all -- although if I'm wrong about that, feel free to open a PR there as well, and link it to this work.

daenney:
Yap, lets close this one and I'll rework that according to this discussion. Thanks!

daenney:
I realised why I suggested the name change, I wanted a way to differentiate between a non-refreshable token having expired and a refreshable token having failed to get refreshed, because that case we don't want to handle magically since there's no way of knowing what to do. But I suppose I can check that based on the offline attribute instead and leave the error in requests-oauthlib for what it is.

louking:
Just curious - is a new PR going to be written to replace this? I was watching this carefully as it is timely for me -- I'm building a new web app and using google for authentication. Much prefer working examples than reinventing the wheel. :)

daenney:
I'm working on a set of changes for it, but I've got some things to do at work that take priority right now. Should hopefully get something working up early next week.

PR #146, opened 2018-07-17
Improve build mechanism

JonathanHuot:
I've added tox file to reproduce the TravisCI build and ReadTheDocs build, **before** pushing changes to GitHub.
At the moment, it does not support the `postgresql` build variant, because it implies to have a postgresql installed on the developer machine. However, we can consider adding it.
Also, the `tox -e docs` is reproducing the exact steps of readthedocs website, i.e. python2.7, and should help when proposing documentation changes in PR.

Why I have added this ? It allows upstream libraries e.g. `oauthlib` to run the `flask-dance` unit tests before releasing new `oauthlib` versions.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,11 @@
+[tox]
+envlist = py27-sqlite{,-blinker},py34-sqlite{,-blinker},py35-sqlite{,-blinker},py36-sqlite{,-blinker}
+
+[testenv]
+deps=
+    -rdev-requirements.txt
+    -rrequirements.txt
+    blinker: blinker
+    codecov
+setenv=DATABASE_URI=sqlite://
+commands=py.test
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -0,0 +1,19 @@
...
+deps=
+    -rdev-requirements.txt
+    -rrequirements.txt
+    blinker: blinker
+    codecov
+setenv=DATABASE_URI=sqlite://
+commands=py.test
+
+[testenv:docs]
+basepython=python2.7
```
singingwolfboy: Could you make this default to Python 3.6 or Python 3.7, please? I'd rather look to the future, rather than stay in the past.

JonathanHuot: While I agree 100% with you, the default settings for ReadTheDocs is to use python2.7. See https://docs.readthedocs.io/en/latest/yaml-config.html#python-version

However, if you change the default of RTD, e.g. creating a `.readthedocs.yml` with: 
```
python:
   version: 3.6
```

Then, it will be natural to change the `basepython` into 3.6 as well in tox.

What do you think ?

singingwolfboy: I already modified the settings for this project on RTD awhile ago, and it builds using Python 3. I wasn't aware that you could use a `.readthedocs.yml` file to customize settings -- I'll check that out. If you want to make one in this pull request, that's fine with me.

singingwolfboy:
Thanks!

PR #151, opened 2018-08-01
google: Update default scope

### LINKED ISSUE #148: Google scope rename
All of a sudden I'm getting bugs in production with the following warning:
```
Warning: Scope has changed from "profile email" to "https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile".
```

It looks like we need to update https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/contrib/google.py#L56 and https://github.com/singingwolfboy/flask-dance/blob/master/docs/quickstarts/google.rst accordingly.

daenney:
Use the full scope. Something's changed on Google's side that now really requires the `OAUTHLIB_RELAX_TOKEN_SCOPE` to be set when using "short" scopes, for some reason that worked before without it. Since this is not extensively documented and the default now causes things to break, update the default to something that works without additional configuration/caveats.

Fixes #148

### DIFF
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -27,7 +27,7 @@ def make_google_blueprint(
         client_id (str): The client ID for your application on Google
         client_secret (str): The client secret for your application on Google
         scope (str, optional): comma-separated list of scopes for the OAuth token.
-            Defaults to the "profile" scope.
+            Defaults to the "https://www.googleapis.com/auth/userinfo.profile" scope.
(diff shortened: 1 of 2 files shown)


singingwolfboy:
I guess we're going to want to cut a new release for this change and push it to PyPI. Is that right? @daenney, do you want to help with that too?

daenney:
Probably should cut a new release, yes. I'm not sure how your release process works, aside from updating the changelog?

PR #153, opened 2018-08-02
Azure tenant endpoints

atwalsh:
Adds a `tenant` argument to `make_azure_blueprint()` to allow different OAuth request endpoints as described here: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints. 


Summary of acceptable tenant values:

| Value                                                               | Description                                                                                                                                                                                                                   |
|---------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `common`                                                            | Allows users with both personal Microsoft accounts and work/school accounts from Azure Active Directory to sign into the application.                                                                                         |
| `organizations`                                                     | Allows only users with work/school accounts from Azure Active Directory to sign into the application.                                                                                                                         |
| `consumers`                                                         | Allows only users with personal Microsoft accounts (MSA) to sign into the application.                                                                                                                                        |
| `8eaef023-2b34-4da1-9baa-8bc8c9d6a490` or `contoso.onmicrosoft.com` | Allows only users with work/school accounts from a particular Azure Active Directory tenant to sign into the application. Either the friendly domain name of the Azure AD tenant or the tenant's guid identifier can be used. |


flask_dance/contrib/azure.py
```diff
@@ -15,7 +15,7 @@
 def make_azure_blueprint(
         client_id=None, client_secret=None, scope=None, redirect_url=None,
         redirect_to=None, login_url=None, authorized_url=None,
-        session_class=None, backend=None):
+        session_class=None, backend=None, tenant=None):
```
singingwolfboy: Instead of defaulting to `None`, can you have this parameter default to "common"? That way, we can take out the `tenant = tenant or "common"` line later on.

I have this construction for the `scope` argument because it's a list, which is a mutable type, and setting a mutable type as a default value for a function argument is a bad idea in Python. But strings are immutable, so it's simpler to just use it as a default value for the argument.

flask_dance/contrib/azure.py
```diff
@@ -41,18 +41,22 @@ def make_azure_blueprint(
         backend: A storage backend class, or an instance of a storage
                 backend class, to use for this blueprint. Defaults to
                 :class:`~flask_dance.consumer.backend.session.SessionBackend`.
+        tenant: Defaults to 'common'. Can be set to one of the four allowed tenant values from:
+                https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints
```
singingwolfboy: Can you use a restructured text link? I suggest something like this:

```
tenant: Determine which accounts are allowed to authenticate with Azure.
        `See the Azure documentation for more information about this parameter.
        <https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints>`_.
        Defaults to ``common``.
```

flask_dance/contrib/azure.py
```diff
@@ -41,18 +41,22 @@ def make_azure_blueprint(
...
     scope = scope or ["user.read"]
     azure_bp = OAuth2ConsumerBlueprint("azure", __name__,
         client_id=client_id,
         client_secret=client_secret,
         scope=scope,
         base_url="https://graph.microsoft.com",
-        authorization_url="https://login.microsoftonline.com/common/oauth2/v2.0/authorize",
-        token_url="https://login.microsoftonline.com/common/oauth2/v2.0/token",
+        authorization_url="https://login.microsoftonline.com/{}/oauth2/v2.0/authorize".format(tenant),
+        token_url="https://login.microsoftonline.com/{}/oauth2/v2.0/token".format(tenant),
```
singingwolfboy: Could you use named format tokens here?
```python
authorization_url="https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize".format(tenant=tenant)
token_url="https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token".format(tenant=tenant)
```

atwalsh:
@singingwolfboy Made requested changes. Thanks!

flask_dance/contrib/azure.py
```diff
@@ -41,6 +41,11 @@ def make_azure_blueprint(
         backend: A storage backend class, or an instance of a storage
                 backend class, to use for this blueprint. Defaults to
                 :class:`~flask_dance.consumer.backend.session.SessionBackend`.
+        tenant: Determine which accounts are allowed to authenticate with Azure.
+                `See the Azure documentation for more information about this parameter.
+                <https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints>`_.
```
singingwolfboy: You've got an extra period at the end of this line

singingwolfboy:
Thanks!

Issue #149, opened 2018-08-01
Azure authorization exception using Flask Dance


leonbrag:
I have following code running in Docker container in Azure:

```
from flask import Flask, redirect, url_for
from flask_dance.contrib.azure import make_azure_blueprint, azure

import time
app = Flask(__name__)



app.secret_key = "supersekrit"
blueprint = make_azure_blueprint(
    client_id="foo",
    client_secret="bar",
)
app.register_blueprint(blueprint, url_prefix="/login")

@app.route('/')
def hello_world():
	if not azure.authorized:
		return redirect(url_for("azure.login"))
	resp = azure.get("/v1.0/me")
	assert resp.ok
	return resp.json()["mail"] + "<br>Hello World! v1 <br> Server's time:<b>" + time.strftime ("%m-%d-%y %I:%M:%S %p")+"</b>"

if __name__ == '__main__':
  app.run()
```

When I navigate to site, I get credential prompt. When credentials are provided, internal server error is reported. From the docker logs:

```
2018-08-01T06:07:42.731887152Z Traceback (most recent call last):
2018-08-01T06:07:42.731891152Z   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
2018-08-01T06:07:42.731934053Z     response = self.full_dispatch_request()
2018-08-01T06:07:42.731937753Z   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
2018-08-01T06:07:42.731941353Z     rv = self.handle_user_exception(e)
2018-08-01T06:07:42.731944653Z   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
2018-08-01T06:07:42.731948153Z     reraise(exc_type, exc_value, tb)
2018-08-01T06:07:42.731951653Z   File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
2018-08-01T06:07:42.731955053Z     raise value
2018-08-01T06:07:42.731958353Z   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
2018-08-01T06:07:42.731961653Z     rv = self.dispatch_request()
2018-08-01T06:07:42.731964853Z   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
[...]

PR #155, opened 2018-08-06
Updated default scopes according to recent moves from Azure.

steven-martins:
Microsoft updated their api to be more "openid" compliant.  ( [v1-v2-comparison](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-compare) )

The graph scope "user.read" has been replaced by "User.Read" or "https://graph.microsoft.com/user.read". 
It is seems enough for personal accounts (like outlook.com), but openid scopes are needed for entreprise accounts ("openid", "profile", "email").
"User.Read" scope is still needed in order to get an access_token.

It is related to singingwolfboy/flask-dance#149 issue.

And I also updated the example in order to display the email in both cases (personal or enterprise account)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Change Log
 unreleased
 ----------
 * Added ``tenant`` argument to ``make_azure_blueprint``
+* Updated Azure AD default scopes. See `issue 149`_.
 
 1.0.0 (2018-06-04)
 ------------------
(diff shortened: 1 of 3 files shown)


singingwolfboy:
@steven-martins: looks great! Please also add an entry to the CHANGELOG file, and then we can get this merged.

steven-martins:
Hello, the changelog entry has been added.

PR #161, opened 2018-08-18
google: Set auto_refresh_url only for offline

daenney:
Only when we request an offline token do we actually get a refresh token that can be used with the `auto_refresh_url` endpoint to refresh the token. As such, when we get an online token instead (the default) we shouldn't set `auto_refresh_url`.

This is the first part of the changes discussed in #143, #144 and just ensures that requests-oauthlib no longer tries to refresh an online token but immediately raises a `TokenExpiredError`. This does not constitute an API change since the resulting error is the same, but would at least result in a more clear traceback.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,9 @@ unreleased
 ----------
 * Added ``tenant`` argument to ``make_azure_blueprint``
 * Updated Azure AD default scopes. See `issue 149`_.
+* Only set ``auto_refresh_url`` in ``make_google_blueprint`` if a token of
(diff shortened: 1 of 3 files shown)


daenney:
Travis is being unhelpful, one of the Python jobs failed due to a Travis-CI error, but it apparently doesn't take it upon itself to restart that job: https://travis-ci.org/singingwolfboy/flask-dance/jobs/417598497

> OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown

Issue #165, opened 2018-08-23
How to override authorization_url parameter?


mamude:
In my application, we need to limit the auth by domain.

How can I change the default:
authorization_url="https://accounts.google.com/o/oauth2/auth"

To:
authorization_url="https://accounts.google.com/o/oauth2/auth?hd=mydomain.com"

Cheers!

singingwolfboy:
There are two options that I can see. The first option is that we can add a new parameter to the `make_google_blueprint()` function to control this. The second option is that you can make your own blueprint using the `OAuth2ConsumerBlueprint` class, and customize it however you want. [Check the file that defines `make_google_blueprint()`: it's pretty straightforward.](https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/contrib/google.py)

Can you try using the `OAuth2ConsumerBlueprint` class directly? If that works for you, can you send a pull request to add a new parameter to the `make_google_blueprint()` function?

mamude:
I was able to successfully implement the first option.
Many thanks for the suggestion.

PR #166, opened 2018-08-23
Add Google hosted domain (hd) url parameter

atwalsh:
Adds a `hosted_domain` argument to `make_google_blueprint`. 

From Google OpenID Connect [docs](https://developers.google.com/identity/protocols/OpenIDConnect#hd-param):
>The hd (hosted domain) parameter streamlines the login process for G Suite hosted accounts. By including the domain of the G Suite user (for example, mycollege.edu), you can indicate that the account selection UI should be optimized for accounts at that domain.

Related to #165.

**Example:**

```python
blueprint = make_google_blueprint(
    client_id="my-key-here",
    client_secret="my-secret-here",
    scope=["profile", "email"],
    hosted_domain="example.com"
)
```

image

### DIFF
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -56,8 +56,8 @@ def make_google_blueprint(
     .. _google_hosted_domain_warning:
     .. warning::
        The ``hosted_domain`` argument **only provides UI optimization**. Don't rely on this argument to control
-       who can access your application. You must verify that the ``hd`` parameter of the response matches the
-       ``hosted_domain`` argument of ``make_google_blueprint``. For example:
+       who can access your application. You must verify that the ``hd`` claim of the response ID token matches the
+       ``hosted_domain`` argument passed to ``make_google_blueprint``. For example:
 
        .. code-block:: python
 
@@ -78,9 +78,9 @@ def make_google_blueprint(
                 resp_json = google.get("/oauth2/v2/userinfo").json()
                 if resp_json["hd"] != blueprint.authorization_url_params["hd"]:
                     google.authenticated = False
-                    requests.post('https://accounts.google.com/o/oauth2/revoke',
-                                  params={'token': token['access_token']},
-                                  headers={'content-type': 'application/x-www-form-urlencoded'})
+                    requests.post("https://accounts.google.com/o/oauth2/revoke",
+                                  params={"token": token["access_token"]},
+                                  headers={"content-type": "application/x-www-form-urlencoded"})
                     session.clear()
                     abort(403)
 
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -77,10 +77,10 @@ def make_google_blueprint(
             def logged_in(blueprint, token):
                 resp_json = google.get("/oauth2/v2/userinfo").json()
                 if resp_json["hd"] != blueprint.authorization_url_params["hd"]:
-                    google.authenticated = False
-                    requests.post("https://accounts.google.com/o/oauth2/revoke",
-                                  params={"token": token["access_token"]},
-                                  headers={"content-type": "application/x-www-form-urlencoded"})
+                    requests.post(
+                        "https://accounts.google.com/o/oauth2/revoke",
+                        params={"token": token["access_token"]}
+                    )
                     session.clear()
                     abort(403)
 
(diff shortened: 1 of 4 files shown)


atwalsh:
> Considering that this feature will set the default domain for users logging in, I would recommend using the name default_domain for this argument.

Makes sense. I agree that the name **`hd`** does not give any context.

However, since we should validate that returned ID token from Google has the correct **`hd`** claim (passing this argument only provides UI optimization, users can still login with different domains), I wonder what your thoughts are on having this argument named differently than the response argument.

```python
google_bp = make_google_blueprint(
    client_id="my-key-here",
    client_secret="my-secret-here",
    scope=["profile", "email"],
    hd="example.com"
)

# Somewhere, probably in `oauth_authorized.connect`
if resp.json()['hd'] != google_bp.authorization_url_params['hd']:
    # user not in domain
```

Do you think it will be confusing to have the blueprint argument named differently than the response from Google? If not, I will change to `default_domain` or `domain_hint`, whichever you think is best. I will add a note to the quickstart about this validation.

daenney:
`hd` stands for [`hosted domain`](https://developers.google.com/identity/protocols/OpenIDConnect#hd-param).

Regardless of what we name it, people will always have to check the `hd` argument if they care for it. That's more a documentaiton issue imho and shouldn't mean we use a completely opaque identifier in our code. It might be prudent to use `hosted_domain` instead of something like `default_domain` though since in that case the link is a bit more obvious (and reasonably findable with the help of a search engine).

However, there also needs to be a pretty big warning in the docs here that this is purely cosmetic/for UI. I don't think the current wording is clear enough about that, which risks creating the assumption in people's minds that flask-dance will enforce this for them, and open their apps up to all kinds of problems.

atwalsh:
`hosted_domain` as the argument name sounds good to me. @singingwolfboy Are you okay with this naming?

I completely agree that there needs to be a clear warning about validating the returned `hd` argument. What about something like this below the `hosted_domain` argument documentation for `make_google_blueprint`, in the quickstart, or both?  

image


Is `@oauth_authorized.connect_via` the right place to do this? If so, how would I go about canceling the "OAuth dance"?

daenney:
The oauth dance will already have completed at this point. Now we're inside the application's own logic. In this case, if the hosted domain doesn't match I would:
* Revoke the current token
* Clear the session
* Raise a 403

flask_dance/contrib/google.py
```diff
@@ -49,6 +49,41 @@ def make_google_blueprint(
...
+                client_secret="bar",
+                scope=["profile", "email"],
+                hosted_domain="example.com"
+            )
+
+            @oauth_authorized.connect_via(google_bp)
+            def logged_in(blueprint, token):
+                resp_json = google.get("/oauth2/v2/userinfo").json()
+                if resp_json["hd"] != blueprint.authorization_url_params["hd"]:
+                    google.authenticated = False
```
singingwolfboy: I'm not sure why you're setting this `authenticated` boolean here. If you're referring to the `google.authorized` check, [that is a property defined by `requests-oauthlib`](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L129), so you can't set it manually. You may want to remove the `access_token` property, though.

flask_dance/contrib/google.py
```diff
@@ -49,6 +49,41 @@ def make_google_blueprint(
...
+                    requests.post("https://accounts.google.com/o/oauth2/revoke",
+                                  params={"token": token["access_token"]},
+                                  headers={"content-type": "application/x-www-form-urlencoded"})
```
singingwolfboy: Is this header actually needed? If so, we should leave it in, but if it's not actually necessary, we should take it out, so that the docs are as simple as possible.

atwalsh: Nope. This header was not needed. Removed in https://github.com/singingwolfboy/flask-dance/pull/166/commits/676176a67edd9db77b09879647cac5db742d90c9.

flask_dance/contrib/google.py
atwalsh: I removed this line in https://github.com/singingwolfboy/flask-dance/pull/166/commits/676176a67edd9db77b09879647cac5db742d90c9. When you say removing the `access_token` do you mean removing this from the `token` variable with something like `del token['access_token']` after the revoke request?

singingwolfboy:
Thanks!

PR #167, opened 2018-08-24
Update Google quickstart docs

awiddersheim:
The scopes provided were a bit misleading compared to what actually is
needed to get up and running.

### DIFF
--- a/docs/quickstarts/google.rst
+++ b/docs/quickstarts/google.rst
@@ -25,7 +25,10 @@ Code
     blueprint = make_google_blueprint(
         client_id="my-key-here",
         client_secret="my-secret-here",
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Can you point me to some documentation about Google scopes that explains why this is necessary?

awiddersheim:
Some of the scopes are documented [here](https://developers.google.com/identity/protocols/googlescopes).

The code itself also has a default `scope` similar to what I'm showing in this PR but for whatever reason isn't reflected in the quickstart guide.

https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/contrib/google.py#L56

PR #168, opened 2018-08-29
travis: Only build pushes to master

daenney:
With the current configuration, any push to any branch will trigger a build. Usually that's unnecesary, as Travis always (separately) also builds the PR.

This changes the configuration such that
* We don't build any branch that gets pushed to except for master
* Any branch for which a PR is opened will automatically be built
* Any branch for which a PR is open that gets new commits pushed to will be built

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,11 @@ env:
   - BLINKER=1 DATABASE_URI=postgresql://localhost/travis_ci_test
   - BLINKER=1 DATABASE_URI=sqlite://
   - BLINKER=0 DATABASE_URI=sqlite://
+
+branches:
+  only:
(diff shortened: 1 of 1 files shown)


daenney:
Grrr Travis

```
3.5 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.5.tar.bz2
$ curl -sSf -o python-3.5.tar.bz2 ${archive_url}
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
Unable to download 3.5 archive. The archive may not exist. Please consider a different version.
```

PR #170, opened 2018-09-05
Initial commit for Zoho provider

tcmRyan:
Creating a custom WebClient that supports the Zoho Oauth standard appears to be the only way to set headers for this provider.

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -181,6 +181,20 @@ Okta
 
 .. _custom-provider:
 
+Zoho
+-------
+.. module:: flask_dance.contrib.zoho
+
+.. autofunction:: make_zoho_blueprint
+
+.. data:: zoho
+
+    A :class:`~werkzeug.local.LocalProxy` to a :class:`requests.Session` that
+    already has the Zoho authentication token loaded (assuming that the user
+    has authenticated with Zoho at some point in the past).
+
+.. _custom-provider:
+
 Custom
 ------
 
--- a/flask_dance/contrib/zoho.py
+++ b/flask_dance/contrib/zoho.py
@@ -1,0 +1,124 @@
+from __future__ import unicode_literals
+from oauthlib.oauth2.rfc6749 import tokens
+from oauthlib.oauth2.rfc6749.clients.web_application import WebApplicationClient
+from flask_dance.consumer import OAuth2ConsumerBlueprint
+from functools import partial
+from flask.globals import LocalProxy, _lookup_app_object
+try:
+    from flask import _app_ctx_stack as stack
+except ImportError:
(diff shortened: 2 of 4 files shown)


docs/providers.rst
```diff
@@ -181,6 +181,20 @@ Okta
 
 .. _custom-provider:
 
+Zoho
+-------
+.. module:: flask_dance.contrib.zoho
+
+.. autofunction:: make_discord_blueprint
```
singingwolfboy: Shouldn't this be `make_zoho_blueprint`?

daenney:
I'd like to see the reprompting for consent supported too. I would like us to use the same argument to the blueprint as the Google one has, to try and keep the same functionality behind the same arguments across blueprints.

Similar to Google, Zoho also supports two different access types, `online`and `offline`, which we should support too. Just like with Google using `access_type=offline` will give you a `refresh_token` in the response that you can store to refresh a user's oauth session and potentially act on their behalf while they're not at their browser.

While reading the docs I noticed something a bit peculiar though:
```
Note: The prompt=consent parameter is mandatory while fetching the Refresh Token.
```

This would mean that if `access_type=offline` is given the first request must include `prompt=consent` too? I'm not quite sure what they're alluding at.

ryan-s:
Fixed the documentation bug and added support for both reprompt and online/offline access following the Google blueprints pattern.  

I haven't noticed any issues with the refresh token yet, but thats probably due to not having been using this provider long enough.  I can dive back into their SDK and see what they do to handle it this weekend.

flask_dance/contrib/zoho.py
```diff
@@ -0,0 +1,124 @@
...
+            for their consent, even if the user has already given their
+            consent. Defaults to False
+     :rtype: :class:`~flask_dance.consumer.OAuth2ConsumerBlueprint`
+    :returns: A :ref:`blueprint <flask:blueprints>` to attach to your Flask app.
+    """
+    scope = scope or ['ZohoCRM.users.all']
+    base_url = "https://www.zohoapis.com/"
+    client = ZohoWebClient(client_id, token_type=ZOHO_TOKEN_HEADER)
+    authorization_url_params = {}
+    authorization_url_params['access_type'] = 'offline' if offline else 'online'
```
singingwolfboy: This construction is a little clearer:
```python
authorization_url_params = {
    "access_type": "offline" if offline else "online"
}
```
Or this, which might be even clearer:
```python
if offline:
    access_type = "offline"
else:
    access_type = "online"
authorization_url_params = {
    "access_type": access_type,
}
```

Ultimately, it's a stylistic thing, and I'll let you decide. 😄

PR #173, opened 2018-09-14
docs: Fix User() model attributes

wsheppard:
This example was ever so slightly broken - works fine with these minor fixes.

### DIFF
--- a/docs/quickstarts/sqla-multiuser.rst
+++ b/docs/quickstarts/sqla-multiuser.rst
@@ -43,6 +43,7 @@ Create a file called ``multi.py`` with the following contents:
     class User(db.Model, UserMixin):
         id = db.Column(db.Integer, primary_key=True)
         username = db.Column(db.String(256), unique=True)
+        email = db.Column(db.String(256), unique=True)
         # ... other columns as needed
 
     class OAuth(OAuthConsumerMixin, db.Model):
@@ -101,7 +102,7 @@ Create a file called ``multi.py`` with the following contents:
                 # Remember that `email` can be None, if the user declines
                 # to publish their email address on GitHub!
                 email=github_info["email"],
-                name=github_info["name"],
+                username=github_info["name"],
             )
             # Associate the new local user account with the OAuth token
             oauth.user = user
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Note that the `User` model in this example includes the comment `# ... other columns as needed`. I actually don't like the idea of specifying these columns -- I would rather that app developers decide which columns _their app_ needs on their user model. If that means that this code can't be copy-pasted and run without modifications, that's perfectly OK with me -- it's meant to be an illustrative example, not a functional application.

However, I'm open to other perspectives. Can you explain why you want to make these changes?

wsheppard:
Totally get that - and I suspected there would be a semi-pseudo-code angle to this.

However, it's not about telling the devs which columns to use, it's that later in the example there is a reference to the "user" and "email" columns in the User() model - neither of which exist in the current code - therefore it is confusing - particularly for a guide on something new which does indeed have some complexity to it.

Apart from that, everything else is indeed _fully_ functional and as the guide does appear to expect users to copy paste this and run it as a test, I think it's a reasonable expectation that it doesn't just throw an exception straight away. 

I am totally new to Flask-Dance, and this was a stumbling block for me as a new user - I had to debug this script before I could understand what Flask-Dance was all about.

W.

singingwolfboy:
Thanks for the explanation! And if this was a stumbling block for you, then we should definitely find a way to make it easier. Kudos for stepping up and sending a pull request, by the way -- I really appreciate you wanting to help out!

What if we add the columns to the User model that are necessary to make this a fully-functional example, but _also_ indicate that these columns aren't actually necessary to make Flask Dance work? I'm thinking something like this:

```python
class User(db.Model, UserMixin):
    id = db.Column(db.Integer, primary_key=True)
    # Your User model can include whatever columns you want: Flask-Dance doesn't care.
    # Here are a few columns you might find useful, but feel free to modify them
    # as your application needs!
    username = db.Column(db.String(256), unique=True)
    email = db.Column(db.String(256), unique=True)
    name = db.Column(db.String(256))
```

That way, the code should run successfully, but hopefully it won't indicate that these columns are actually _required_.

singingwolfboy:
Yep, that's why I added both an `email` column and a `name` column. Do you want to update this pull request to make that change?

wsheppard:
Sure OK will do.

wsheppard:
OK done - sorry it's multiple commits I made a mistake. If you want to keep history in tact without doing a squash-before-merge, I can resubmit as a single commit. ( I know it's terrible but I'm using the GitHUB editor! First time for everything... )

PR #174, opened 2018-09-14
docs:sqla: Allow functionality out-of-the-box

wsheppard:
Allow the sqlalchemy example to run out-of-the-box without modifications but clarify that model attributes are entirely up to the end-user and not Flask-Dance specific.


wsheppard:
See also #173

singingwolfboy:
Thanks!

Issue #145, opened 2018-07-13
How to logout - unauthorize?


Darthholi:
All the examples mention how to add blueprint, check for authorization and authorize if needed. Is there any way to programatically (in my own logout function) clear the session to let user login(authorize) again?
Thanks!

daenney:
Yes, you can absolutely do this. What you need to do is delete the token from the session. If you're not using the SQLAlchemy backend but going through Flask and it's default cookie-based session storage, you can do something like: `del current_app.blueprints['google'].token`.

Depending on what you want you to achieve you can also revoke the token itself, though this is provider dependent. If you do that they'll be presented with a consent screen next time they login.

Finally, after having removed the token you could also call a `session.clear()` to clear anything else in your session.

daenney:
One thing, if you do this but you use something like a `@login_required` decorator on every page, including `/` the user will automatically get logged back in again as it'll trigger the OAuth flow, they've already given consent and done. In that case you'd have to revoke the token first.

It's also important to note that if they've already given consent but you want them to be able to switch account (perhaps they have multiple and they picked the wrong one) just deleting the token won't work, you'll have to call the OAuth provider's token revocation endpoint first, or force them to go through the consent screen every time.

daenney:
I'm going to close this issue as we haven't heard back in a while. If you need further help with this feel free to re-open the issue.

ohenrik:
I struggled a bit with this as i did not find any clear examples of how to revoke the token from google. However i managed to get it working with this code:

```python
@app.route("/logout")
def logout():
    resp = google.post(
        'https://accounts.google.com/o/oauth2/revoke',
        params={'token': app.blueprints['google'].token["access_token"]},
        headers = {'content-type': 'application/x-www-form-urlencoded'}
    )
    if resp.ok:
        del app.blueprints['google'].token
        session.clear()
        logout_user()
    return redirect(url_for("index"))
```

@daenney Please let us know if there are better ways to implement this logout logic (Especially if there are any security holes in this logic).

daenney:
You could check if you're `google.authorized` first. If your token has expired for example the call to `google.post` will fail with something like an `InvalidClentError` when using online tokens.

`session.clear()` will already remove the token as it nukes everything in the session, so there's no need for the `del` before it.

Other than that, looks good!

ohenrik:
Perfect! Thank you for answering! :))

hammer2383:
` resp = google.post(
[...]

PR #177, opened 2018-10-09
Add documentation for how to log out a user

singingwolfboy:
Since logging out seems to be a problem for several people (see #145), here's some documentation describing the options.

@daenney, can you review this and let me know it it's understandable? Is there anything I'm missing?

### DIFF
--- a/docs/logout.rst
+++ b/docs/logout.rst
@@ -125,6 +125,13 @@ Here's some sample code that works with Google:
 After the user uses this method to log out, Google will not remember that they
 granted consent to share information with your website.
 
+.. warning::
+
+    In this sample code, we are using an :keyword:`assert` statement.
+    This works fine for debugging, but not for production. Be sure to modify
+    this code to appropriately handle cases where there is an API failure
+    when trying to revoke the token.
+
 .. note::
 
     In this code, we already have a reference to the ``blueprint`` object,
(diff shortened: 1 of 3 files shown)


docs/logout.rst
```diff
@@ -0,0 +1,162 @@
...
+
+    @app.route("/logout")
+    def logout():
+        token = blueprint.token["access_token"]
+        resp = google.post(
+            "https://accounts.google.com/o/oauth2/revoke",
+            params={"token": token},
+            headers={"Content-Type": "application/x-www-form-urlencoded"}
+        )
+        assert resp.ok, resp.text
```
daenney: I'm a bit worried about this example. `assert` gets compiled out if you'd start python with `-O` which people might do in production.

I'm also not sure I'd want to break if revoking the access grant failed, in this case that would raise an `AssertionError` that's not handled resulting in a internal server error instead. I'm fine with keeping it as an example but I think it's worth documenting that in reality you should probably handle the case of `!resp.ok` in some gracefull manner instead.

daenney:
I think it's complete and accessible, this should prove useful! I've left one comment related to the example code but other than that, :+1: :100:!

docs/logout.rst
singingwolfboy: Good point. I added a warning about the `assert` statement. Do you think that's enough?

PR #181, opened 2018-10-20
Add Reddit pre-set configuration

r8:
I'm using Flask Dance for my Reddit-related project, so decided to extract my custom Reddit conf.

Unfortunately, I'm pretty new to pytest and oauthlib, so I'm not sure how to achieve 100% coverage.
It seems that one statement in my custom session is still missed.

### DIFF
--- a/flask_dance/contrib/reddit.py
+++ b/flask_dance/contrib/reddit.py
@@ -1,0 +1,107 @@
+from __future__ import unicode_literals
+
+from flask_dance.consumer import OAuth2ConsumerBlueprint, OAuth2Session
+from functools import partial
+from flask.globals import LocalProxy, _lookup_app_object
+
+from flask_dance import __version__ as _flask_dance_version
+
+try:
+    from flask import _app_ctx_stack as stack
+except ImportError:
+    from flask import _request_ctx_stack as stack
+
+__maintainer__ = "Sergey Storchay <r8@r8.com.ua>"
+
+
+def _default_user_agent():
+    return "Flask-Dance/%s" % _flask_dance_version
+
+
+class RedditOAuth2Session(OAuth2Session):
+    def __init__(self, *args, **kwargs):
+        super(RedditOAuth2Session, self).__init__(*args, **kwargs)
+
+        # Reddit requires to set non-generic user agent
+        self.headers["User-Agent"] = self.blueprint.user_agent or _default_user_agent()
+
+    def fetch_token(self, *args, **kwargs):
+        # Pass client_id to session so it could trigger Basic Auth
+        return super(RedditOAuth2Session, self).fetch_token(
+            client_id=self.blueprint.client_id, *args, **kwargs)
+
+
+def make_reddit_blueprint(user_agent=None,
+        client_id=None, client_secret=None, scope="identity", permanent=False,
+        redirect_url=None, redirect_to=None, login_url=None, authorized_url=None,
+        session_class=None, backend=None):
+    """
+    Make a blueprint for authenticating with Reddit using OAuth 2. This requires
+    a client ID and client secret from Reddit. You should either pass them to
+    this constructor, or make sure that your Flask application config defines
+    them, using the variables REDDIT_OAUTH_CLIENT_ID and REDDIT_OAUTH_CLIENT_SECRET.
+
+    Args:
+        user_agent (str, optional): User agent for the requests to Reddit API.
+            Defaults to ``Flask-Dance/{{version}}``
+        client_id (str): The client ID for your application on Reddit.
+        client_secret (str): The client secret for your application on Reddit
+        scope (str, optional): space-separated list of scopes for the OAuth token
+            Defaults to ``identity``
+        permanent (bool, optional): Whether to request permanent access token.
+            Defaults to False, access will be valid for 1 hour
+        redirect_url (str): the URL to redirect to after the authentication
+            dance is complete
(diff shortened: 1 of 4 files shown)


flask_dance/contrib/reddit.py
```diff
@@ -0,0 +1,107 @@
...
+
+def _default_user_agent():
+    return "Flask-Dance/%s" % _flask_dance_version
+
+
+class RedditOAuth2Session(OAuth2Session):
+    def __init__(self, *args, **kwargs):
+        super(RedditOAuth2Session, self).__init__(*args, **kwargs)
+
+        # Reddit requires to set non-generic user agent
```
singingwolfboy: Grammar fix:

```suggestion
        # The Reddit API requires a non-generic user agent
```

flask_dance/contrib/reddit.py
```diff
@@ -0,0 +1,107 @@
...
+        self.headers["User-Agent"] = self.blueprint.user_agent or _default_user_agent()
+
+    def fetch_token(self, *args, **kwargs):
+        # Pass client_id to session so it could trigger Basic Auth
+        return super(RedditOAuth2Session, self).fetch_token(
+            client_id=self.blueprint.client_id, *args, **kwargs)
+
+
+def make_reddit_blueprint(user_agent=None,
```
singingwolfboy: Please put the `user_agent` argument at the *end* of the list of arguments, not at the start. Every other configuration has `client_id` and `client_secret` as the first two arguments, so that the developer can pass them as positional arguments. The Reddit configuration should match that convention.

flask_dance/contrib/reddit.py
```diff
@@ -0,0 +1,107 @@
...
+        redirect_url=None, redirect_to=None, login_url=None, authorized_url=None,
+        session_class=None, backend=None):
+    """
+    Make a blueprint for authenticating with Reddit using OAuth 2. This requires
+    a client ID and client secret from Reddit. You should either pass them to
+    this constructor, or make sure that your Flask application config defines
+    them, using the variables REDDIT_OAUTH_CLIENT_ID and REDDIT_OAUTH_CLIENT_SECRET.
+
+    Args:
+        user_agent (str, optional): User agent for the requests to Reddit API.
```
singingwolfboy: Similarly, please move this documentation to the end of the list of arguments.

tests/contrib/test_reddit.py
```diff
@@ -0,0 +1,96 @@
...
+    assert client_id == "foo"
+
+
+@responses.activate
+def test_context_local():
+    responses.add(responses.GET, "https://google.com")
+
+    # set up two apps with two different set of auth tokens
+    app1 = Flask(__name__)
+    ghbp1 = make_reddit_blueprint(
```
singingwolfboy: Could you change these variables to `reddit_bp1` and `reddit_bp2`?

flask_dance/contrib/reddit.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self, *args, **kwargs):
+        super(RedditOAuth2Session, self).__init__(*args, **kwargs)
+
+        # Reddit requires to set non-generic user agent
+        self.headers["User-Agent"] = self.blueprint.user_agent or _default_user_agent()
+
+    def fetch_token(self, *args, **kwargs):
+        # Pass client_id to session so it could trigger Basic Auth
+        return super(RedditOAuth2Session, self).fetch_token(
+            client_id=self.blueprint.client_id, *args, **kwargs)
```
singingwolfboy: Can you explain why this is necessary? I'm not familiar with the Reddit API, so I don't know what their requirements are.

r8:
Thanks for your feedback!
I'll update the code soon.

flask_dance/contrib/reddit.py
r8: Again, Flask (and Python in general) is not my usual stack, so I might be missing something or using non-optimal way to do the things.

Reddit fetch token API is protected by Basic Auth. I checked `requests_oauthlib` and saw that Basic Auth can be triggered by passing user/password or just client_id to `fetch_token` method.

https://github.com/requests/requests-oauthlib/blob/c61c6fbe19908df5b8bf5d8caacde3448f2e0c49/requests_oauthlib/oauth2_session.py#L201

Obvious way would be to pass it with `token_url_params` through `OAuth2ConsumerBlueprint` constructor.
But if we aren't passing client_id/client_secret to `make_reddit_blueprint` and instead are taking it from config, then we don't know client_id during `OAuth2ConsumerBlueprint` initialization.

Hope that makes sense.

So the easiest way for me was to override `fetch_token` and take client_id from Blueprint. Because at the moment of `fetch_token` call we should know for sure client_id (either from initialization or Flask config).

It was a quick hack, so I'm open for suggestions about more elegant solution :)

singingwolfboy:
Sorry it took me so long to get back to this, but now it looks great! Thanks for making the improvements I suggested. 😄 I'm not bothered by missing one line of test coverage -- maybe I'll clean it up in the future. For now, I think this is ready to be merged. 🎉

r8:
Thank you :)

singingwolfboy:
FYI, this is now released in [Flask-Dance 1.2.0](https://pypi.org/project/Flask-Dance/1.2.0/)

r8:
Thanks!
I forgot to add it to the list in README though

singingwolfboy:
@r8 I just created #194, which removes the provider list from the README. Do you think it's a good change?

PR #185, opened 2018-11-15
Add option for facebook blueprint to rerequest authorization

Tadaboody:
Prior to this change, if an existing user tried to login after refusing
an authorization scope, the dialog would just redirect him to sucess
without asking again for the missing scopes.

Sarting from this commit, using a flag and as documented
[here](https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#reaskperms) a blueprint can be set to re-request the declined scopes

### DIFF
--- a/tests/contrib/test_facebook.py
+++ b/tests/contrib/test_facebook.py
@@ -99,5 +99,7 @@ def test_rerequest_declined_scopes(rerequest):
         )
     assert resp.status_code == 302
     location = URLObject(resp.headers["Location"])
-    # Using dict.get because the "auth_type" header might not be set - this is valid behaviour
-    assert (location.query_dict.get("auth_type") == "rerequest") == rerequest
+    if rerequest:
+        assert location.query_dict["auth_type"] == "rerequest"
+    else:
+        assert "auth_type" not in location.query_dict
--- a/tests/contrib/test_facebook.py
+++ b/tests/contrib/test_facebook.py
@@ -80,6 +80,12 @@ def test_context_local():
 
 @pytest.mark.parametrize("rerequest", (True, False))
 def test_rerequest_declined_scopes(rerequest):
+    """
+    Tests that the rerequest_declined_permissions flag in the facebook blueprint sends
+    toggles the header reasking oauth permissions as detailed in the facebook docs https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#reaskperms
+
+    Tests both that the header is set with the flag (rerequest=True) and that it is missing without the flag (rerequest=False)
+    """
     app = Flask(__name__)
     app.secret_key = "backups"
     bp = make_facebook_blueprint(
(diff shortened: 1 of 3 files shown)


Tadaboody:
Blunder fixed - Tests coming soon™️

singingwolfboy:
@Tadaboody any idea when you'll have time to write those tests?

Tadaboody:
@singingwolfboy Done! split the parametrization into a seprate test in case you prefer the more straightforward (but way less DRY) approach.
I was first worried that the failure logs won't be informative if I go with this approach but pytest does it yet again! these are the failure logs when negating equality operator (==->!=)
```
>       assert (location.query_dict.get("auth_type") == "rerequest") != rerequest
E       AssertionError: assert (None == 'rerequest') != False
E        +  where None = <built-in method get of dict object at 0x113118cf0>('auth_type')
E        +    where <built-in method get of dict object at 0x113118cf0> = {'client_id': 'None', 'redirect_uri': 'https://a.b.c/facebook/authorized', 'response_type': 'code', 'scope': 'user_posts', ...}.get
E        +      where {'client_id': 'None', 'redirect_uri': 'https://a.b.c/facebook/authorized', 'response_type': 'code', 'scope': 'user_posts', ...} = URLObject('https://www.facebook.com/dialog/oauth?response_type=code&client_id=None&redirect_uri=https%3A%2F%2Fa.b.c%2Ffacebook%2Fauthorized&scope=user_posts&state=ItuFsOH1mirGhVZSbJ5366TZtyLGNw').query_dict
```

tests/contrib/test_facebook.py
```diff
@@ -76,3 +76,28 @@ def test_context_local():
...
+    with app.test_client() as client:
+        resp = client.get(
+            "/facebook",
+            base_url="https://a.b.c",
+            follow_redirects=False,
+        )
+    assert resp.status_code == 302
+    location = URLObject(resp.headers["Location"])
+    # Using dict.get because the "auth_type" header might not be set - this is valid behaviour
+    assert (location.query_dict.get("auth_type") == "rerequest") == rerequest
```
singingwolfboy: This line is confusing: it's trying to do too much. Can you break it up like this instead?

```python
if rerequest:
    assert location.query_dict["auth_type"] == "rerequest"
else:
    assert "auth_type" not in location.query_dict
```

singingwolfboy:
Also, please add something to the `CHANGELOG.rst` file to document this change. You can put it under the "unreleased" section.

Tadaboody:
Done and done, I'd add a checklist in the PR template that mentions writing tests and adding the changes to the CHANGELOG so you don't have to ask every time. kind of like what [pytest has](https://github.com/pytest-dev/pytest/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

singingwolfboy:
Fantastic! Thank you for getting this over the finish line 🏁

PR #187, opened 2018-11-22
added blueprint_name default parameter

sreevardhanreddi:
I have used flask_dance for a project and noticed that we cannot use "make_google_blueprint" function in more than one file, as the name of the blueprint is default to "google" in google.py file ( there cannot be same files with the same blueprint name ), I have added a Keyword argument in the "make_google_blueprint" function as blueprint_name=" google". So while creating a google_blueprint, in more than one file, the blueprint_name can be set accordingly, to avoid same blueprint name error

### DIFF
--- a/flask_dance/contrib/azure.py
+++ b/flask_dance/contrib/azure.py
@@ -15,7 +15,7 @@ __maintainer__ = "Steven MARTINS <steven.martins.fr@gmail.com>"
 def make_azure_blueprint(
         client_id=None, client_secret=None, scope=None, redirect_url=None,
         redirect_to=None, login_url=None, authorized_url=None,
-        session_class=None, backend=None, tenant="common"):
+        session_class=None, backend=None, tenant="common", blueprint_name="azure"):
     """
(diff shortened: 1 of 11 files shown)


singingwolfboy:
I don't particularly like this solution. If you need to drastically customize the blueprint returned from `make_google_blueprint()` (or any other function like that), then it would be better to not use the function at all. Just import the `OAuth2ConsumerBlueprint` class and instantiate it directly.

Is there a reason why you can't do that?

daenney:
I believe the idea is that they want to create multiple blueprints, with their defaults, but named differently? I supposed b/c they're configured with different Oauth configurations?

Calling `make_google_blueprint` means it always creates and attaches a blueprint named `google`, which we can then access through `current_app.blueprints['google']`.

I guess if you have two sets of routes in your application you might want to have two Google oauth providers configured, fronting the admin route with one oauth client and the regular routes with a different one (perhaps one configured to only allow specific domains vs everyone)?

PR #189, opened 2018-11-24
Fixed the formating so the first args are formated correctly

tcmRyan:
Blank line ensures the docstrings get parsed correctly by sphinx.


singingwolfboy:
Thanks!

PR #193, opened 2018-12-03
Pin oauthlib to avoid breakage with 3.0.0 release

skion:
We're going to release OAuthLib 3.0.0 soon, which contains some breaking changes.

This PR is to avoid problems before we are able to address the compatibility.


daenney:
Thanks for this, that's really kind to reach out to your consumers! I guess we'll also need to pin a version range for requests-oauthlib once requests/requests-oauthlib#341 gets merged.

PR #195, opened 2018-12-12
Add option to allow CSRF attacks

### LINKED ISSUE #191: Using Flask-dance for Slack app directory installation or with Add To Slack button
Hi @singingwolfboy!
Thank you for the great Flask app to manage OAuth dance.

I use Flask-Dance to manage Slack authentication and ran into an issue where your experience may be beneficial.
If you build OAuth dance starting from `/login/slack` view, everything works fine. However, if you try to install the app from _Add to Slack_ button or directly from Slack app directory, it forwards you to 
`https://slack.com/oauth/authorize?scope=SCOPE&client_id=CLIENT_ID`, and once you authorize the app, Slack redirects you to a `redirect_url` that I set in my Slack app to `login/slack/authorized`.
Slack sending the request to `login/slack/authorized` does not set the state as the OAuth dance is not initiated from the consumer.
Here comes the problem. As per `flask_dance/consumer/oauth2.py`, lines 225-228:
```
        if state_key not in flask.session:
            # can't validate state, so redirect back to login view
            log.info("state not found, redirecting user to login")
            return redirect(url_for(".login"))
```
So, Flask-Dance does not see the state parameter recorded in the current session and asks Slack to authorize the app for the second time, now sending the request with `state` parameter set, and the user has to click _Authorize_ twice.

It's probably not a problem of Flask-Dance per se, but I would not like to implement a separate Slack redirect URL just for this install flow.

What do you think, how can work around this problem? Or maybe you could suggest a pull request to Flask-Dance?

Thanks!

singingwolfboy:
Fixes #191. Slack apps can be installed from the [Slack app directory](https://slack.com/apps), which involves doing the OAuth dance starting from slack.com instead of from the Flask app. This is the same as a [cross-site request forgery attack](https://en.wikipedia.org/wiki/Cross-site_request_forgery), but it is the expected behavior.

Is there some way that we can narrow the scope of this vulnerability? Is there a reliable way to _only_ bypass the `state` check for requests that were initiated by a subdomain of `slack.com`, for example? I don't know if `Referer` headers are reliable or not...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ Changelog
 `unreleased`_
 -------------
 
-nothing yet
+* Added option to allow CSRF attacks, needed for Slack apps.
+  See `#191`_.
 
 `1.2.0`_ (2018-12-05)
 ---------------------
@@ -259,6 +260,7 @@ Fixed
 .. _#143: https://github.com/singingwolfboy/flask-dance/issues/143
 .. _#144: https://github.com/singingwolfboy/flask-dance/issues/144
 .. _#161: https://github.com/singingwolfboy/flask-dance/issues/161
+.. _#191: https://github.com/singingwolfboy/flask-dance/issues/191
 
 
 .. _unreleased: https://github.com/singingwolfboy/flask-dance/compare/v1.2.0...HEAD
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -40,6 +40,7 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             token_url_params=None,
             redirect_url=None,
             redirect_to=None,
+            allow_csrf=False,
             session_class=None,
             backend=None,
 
@@ -93,6 +94,15 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
                 :func:`~flask.url_for` with this argument. If you do not specify
                 either ``redirect_url`` or ``redirect_to``, the user will be
                 redirected to the root path (``/``).
+            allow_csrf (bool): Normally, the OAuth dance uses a ``state`` token
+                to prevent `cross-site request forgery (CSRF) attacks
+                <https://en.wikipedia.org/wiki/Cross-site_request_forgery>`_.
+                However, some OAuth providers expect a third-party website to
+                initiate the OAuth dance on behalf of the consumer,
+                even though this is indistinguishable from a CSRF attack.
+                Set this option to ``True`` in order to allow *both*
+                requests from third-party websites, *and* CSRF attacks.
+                Defaults to ``False``.
             session_class: The class to use for creating a Requests session
                 between the consumer (your website) and the provider (e.g.
                 Twitter). Defaults to
@@ -133,6 +143,7 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
         self.token_url_params = token_url_params or {}
         self.redirect_url = redirect_url
         self.redirect_to = redirect_to
+        self.allow_csrf = allow_csrf
 
         self.teardown_app_request(self.teardown_session)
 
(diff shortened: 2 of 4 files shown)


singingwolfboy:
@Lexy2, @daenney, what do you two think of this?

remyabel:
>  Is there a reliable way to only bypass the state check for requests that were initiated by a subdomain of slack.com, for example? I don't know if Referer headers are reliable or not...

The slack docs has instructions on [verifying requests from slack to the app](https://api.slack.com/docs/verifying-requests-from-slack).

daenney:
I worry about having this as a global on the provider. I don't like the idea of this potentially weakening the Oauth flow for everything when there's only 1 very specific case we want this.

I would prefer to see this implemented by first validating the Signing header as coming from Slack (which requires people to pass in the app's signing key) and only if that passes have the request honour the `allow_csrf`.

Lexy2:
Totally agree with @daenney. Weakening the provider globally is not what @singingwolfboy suggested in #191. Having a separate url route for app directory installation looks better.
Verifying Slack signing headers would not work because Slack does not sign requests with app's signing secret in OAuth flow (i.e. until the app is authorized by the user).

singingwolfboy:
At first, I tried to write a separate URL route. However, I realized that I essentially would need to copy the code from the `authorized` route, and maintaining duplicate code is generally a bad idea. The only way I saw to get around the duplicate code is to have a boolean flag to switch the `state` checking on and off, so that's what I did here.

As far as weakening the OAuth flow for everything, I specifically did _not_ add `allow_csrf` arguments for the `make_X_blueprint()` factory functions that already exist for other providers. It's only there for the Slack blueprint. That means that if you're using one of these factory functions (other than Slack), you *can't* enable CSRF attacks. You have to use the `OAuth2ConsumerBlueprint` class directly to get access to that flag, which seems like a reasonable trade-off to me.

If there's a better way to implement this, I'm happy to hear this! That's the reason why I haven't merged this pull request yet; it's the best implementation I can think of, but I'm open to other possible implementations instead.

daenney:
I think we're getting our hairs crossed. What I meant to say with "weakening the OAuth flow for everything" was "weakening the OAuth flow for every authentication using the Slack provider".

My point being the following; even with Slack there is only 1 very specific case in which we want to allow the CSRF "attack", that is when the OAuth flow is initiated by the Slack App Directory installation. In every other case, we don't really want the CSRF possibility and I'm hoping we can strike a better balance.

So, instead of having a "global" flag on the Oauth2 blueprint I'm hoping we can have a function/request scoped value instead where we can explicitly turn this on for a request, but not all requests.

The idea being the following: if we're getting an Oauth2 flow for Slack, check if there's a state parameter:
- If state is present, do the regular thing
- If state is not present, check for the `X-Slack-Signature` header:
  - If present, validate the signature according to https://api.slack.com/docs/verifying-requests-from-slack and should the signature be valid only then allow it to proceed without the state. This will require a kwarg on `make_slack_blueprint` to let people pass in their Slack Signing Secret
  - If not present or validation fails, abort

Lexy2:
@daenney,
> If state is not present, check for the X-Slack-Signature header:
Slack does not send this header during OAuth dance. X-Slack-Signature can only be used after one has authorized the app.

daenney:
Oh. Damn. THen I've misunderstood that completely. In that case, ignore everything I said.

singingwolfboy:
OK, it sounds like the current state of this pull request is:

* @Lexy2 is taking a close look at the code
* @daenney thought there was a mitigation for the problem, but turned out to be wrong

Is that correct? @daenney, are you OK with this code being merged as-is? I'm going to wait a few days for @Lexy2 to post another comment after investigating the code in more detail. However, if I don't hear anything further, or if @Lexy2 doesn't have any suggested changes, then I'm going to merge this PR. Everyone alright with that?

Lexy2:
I'm on it, David.

I don't want this PR merged in the current state.

daenney:
@singingwolfboy Correct

singingwolfboy:
Based on the discussion in #201, it seems like this feature (or security hole) isn't necessary after all, so I'm closing this pull request.

PR #196, opened 2018-12-12
Use public oauthlib method

singingwolfboy:
https://github.com/oauthlib/oauthlib/pull/546


Tadaboody:
This isn't detailed in the changelog. 
I find it important enough because this change fixes 10s of warnings in my code, and I wasn't sure if it was fixed

PR #200, opened 2018-12-18
Add pre-set configuration for Authentiq

skion:
This adds pre-configured support for [Authentiq](https://www.authentiq.com/), and also sorts the longish list of providers.

### DIFF
--- a/flask_dance/contrib/authentiq.py
+++ b/flask_dance/contrib/authentiq.py
@@ -16,9 +16,10 @@ __maintainer__ = "Pieter Ennes <support@authentiq.com>"
 
 
 def make_authentiq_blueprint(
-        client_id=None, client_secret=None, scope=None, redirect_url=None,
-        redirect_to=None, login_url=None, authorized_url=None,
-        session_class=None, backend=None, hostname="connect.authentiq.io"):
+        client_id=None, client_secret=None, scope="openid profile",
+        redirect_url=None, redirect_to=None, login_url=None,
+        authorized_url=None, session_class=None, backend=None,
+        hostname="connect.authentiq.io"):
     """
     Make a blueprint for authenticating with authentiq using OAuth 2. This requires
     a client ID and client secret from authentiq. You should either pass them to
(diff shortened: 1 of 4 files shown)


singingwolfboy:
Sorting the provider list is a good idea -- I've cherry-picked that commit into master. If you rebase, that commit should disappear from your pull request. You'll also need to fix the test failures.

tests/contrib/test_authentiq.py
```diff
@@ -0,0 +1,97 @@
...
+from urlobject import URLObject
+
+from flask_dance.consumer import OAuth2ConsumerBlueprint
+from flask_dance.consumer.backend import MemoryBackend
+from flask_dance.contrib.authentiq import make_authentiq_blueprint, authentiq
+
+
+def test_blueprint_factory_default():
+    # Test with connect.authentiq.io
+    glbp1 = make_authentiq_blueprint(
```
singingwolfboy: Can you name this variable `abp1` instead? (Short for Authentiq blueprint 1). You probably copied this from the GitLab automated tests, which is fine, but please make sure the variables reflect the content of the tests, here and throughout this file.

flask_dance/contrib/authentiq.py
```diff
@@ -0,0 +1,84 @@
...
+                backend class, to use for this blueprint. Defaults to
+                :class:`~flask_dance.consumer.backend.session.SessionBackend`.
+        hostname (str, optional): If using a private instance of authentiq CE/EE,
+            specify the hostname, default is ``connect.authentiq.io``
+
+    :rtype: :class:`~flask_dance.consumer.OAuth2ConsumerBlueprint`
+    :returns: A :ref:`blueprint <flask:blueprints>` to attach to your Flask app.
+    """
+    scope = set(["openid"] + (scope or "").split())
+    scope = " ".join(scope)
```
singingwolfboy: I think you're trying to be too fancy here. I recommend that you just set `scope="openid"` in the function definition, so that "openid" is the default value. Let the developer decide what the scope should be, and pass that value through without modification.

tests/contrib/test_authentiq.py
skion: Changed to `aqbp`, since that is the common abbreviation used.

skion:
@skion Thanks, should be better now!

flask_dance/contrib/authentiq.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import unicode_literals
+
+from functools import partial
+
+from flask.globals import LocalProxy, _lookup_app_object
+
```
singingwolfboy: Any particular reason for all the blank lines here? It's just a problem, but it looks a bit weird, especially since none of the other configurations are like this.

daenney:
I've merged it since everything looked OK and all points of review were addressed from what I could see.

PR #201, opened 2018-12-22
Allow weak route with no CSRF protection

Lexy2:
Hi, @daenney and @singingwolfboy,

Kindly review the proposed changes.

I have reworked your branch and instead of globally weakening security for Slack, I added an `enable_slack_app_directory` option to `make_slack_blueprint` function and `SlackBlueprint` blueprint. If this parameter is set to True, `SlackBlueprint` creates both weak and strong OAuth routes - `/login/slack/authorized` and `/login/slack/authorized/strong`.

I wanted to make `/login/slack/weak_authorized`, but [Slack docs](https://api.slack.com/docs/oauth#redirect_urls) say that 
```
The redirect URL's path must reference a subdirectory of the callback URL
```
So I had to do it this way.

If `enable_slack_app_directory` is omitted or set to False, the blueprint functions the same way as before.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 `unreleased`_
 -------------
 
-* Added option to allow CSRF attacks, needed for Slack apps.
+* Added option to disable CSRF protection, needed for Slack app installation from app directory.
   See `#191`_.
 
 `1.2.0`_ (2018-12-05)
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -40,7 +40,6 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             token_url_params=None,
             redirect_url=None,
             redirect_to=None,
-            allow_csrf=False,
             session_class=None,
             backend=None,
 
@@ -48,7 +47,7 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
         """
         Most of the constructor arguments are forwarded either to the
         :class:`flask.Blueprint` constructor or the
-        :class:`requests_oauthlib.OAuth2Session` construtor, including
+        :class:`requests_oauthlib.OAuth2Session` constructor, including
         ``**kwargs`` (which is forwarded to
         :class:`~requests_oauthlib.OAuth2Session`).
         Only the arguments that are relevant to Flask-Dance are documented here.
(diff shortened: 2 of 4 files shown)


Lexy2:
Fixes #191

daenney:
Could we flip it? `/login/slack/authorized` is currently in used by the "strong" authentication, changing it to be weak is essentially a backwards incompatible change. So I would prefer to see `/login/slack/authorized/weak` instead.

I also find the documentation confusing:

```
            Defaults to ``/slack/authorized``. If ``enable_slack_app_directory``
            is set to `True`, defaults to ``/slack/authorized/strong``
```

Right now, setting `enable_slack_app_directory` seems to suggest it strengthens the authentication flow, which is quite the opposite.

Lexy2:
Well, there is one more possibility. A Slack app can have both `/login/slack/authorized/weak` and `/login/slack/authorized` redirect URIs. 
I'll need to test if this works properly.

singingwolfboy:
This seems more complicated than it needs to be, and apparently doesn't work properly, either. Seems like #195 is better, so if we need this feature, I think we should go back to that PR. What are your thoughts?

Lexy2:
David, can you elaborate, please? I think I'm misunderstanding something.
> apparently doesn't work properly, either

What exactly doesn't work properly?

> Seems like #195 is better

#195 disables CSRF protection completely for the whole dance, regardless of its source.
This one disables CSRF protection only for Slack app directory source, keeping app-initiated authentication secure.
The code is not that complicated to me.

What are your thoughts?

singingwolfboy:
> What exactly doesn't work properly?

I'm referring to [your comment above](https://github.com/singingwolfboy/flask-dance/pull/201#issuecomment-450756460), where you said "I'll need to test if this works properly", and then "Nah, doesn't work this way". Maybe you were referring to something different?

> This one disables CSRF protection only for Slack app directory source, keeping app-initiated authentication secure.

Unfortunately, that's not the way security works. Imagine that you are an attacker, trying to exploit the OAuth authentication system for a website that uses Flask-Dance. You know that there's two possible URLs you can target: `/login/slack/authorized/weak` and `/login/slack/authorized`. Which one are you going to try to attack? The weaker one, of course.

Exposing a higher-security URL does *not* increase security if you also expose a weak URL. If you're going to expose the weak URL, you might as well not bother with the higher-security URL at all. That's why #195 disables CSRF protection completely: there's no reason not to, and it makes the code simpler.

> The code is not that complicated to me.

Of course, because you wrote it. But try to imagine someone approaching this codebase for the first time, and trying to understand and debug this code. You have to understand how the `authorized_url` and `weak_authorized_url` variables are set and modified, then understand what `OAuth2ConsumerBlueprint.__init__()` does (as well as the subtleties of calling `.__init__()` vs `super()`), understand how the `partial` function works, and understand what makes the weak authorization "weaker" than the regular authorization.

Lexy2:
> Maybe you were referring to something different?

Yes, I was referring to the route pair of `/login/slack/authorized` and `/login/slack/authorized/weak` that doesn't work. The route pair of `/login/slack/authorized` and `/login/slack/authorized/strong` that is in the PR works.

> Unfortunately, that's not the way security works. <...>
> Exposing a higher-security URL does not increase security if you also expose a weak URL.

And you are very right here, thank you for outlining this scenario for me — it's the deal-breaker.
I don't want to weaken the security just to save one click for development applications.

Production applications (published to Slack app directory) can have a [direct install URL](https://api.slack.com/slack-apps#direct_install) that facilitates installation without compromising the security.

In this case, there is no need to implement a weaker route in Flask-Dance at all. I will set the direct install URL to `<host URI>/login/slack` - and the app directory will get the state parameter as required by the RFC.

Thank you for your time and patience.

Regards, Lexy

PR #202, opened 2018-12-31
Update to the twitter quickstart about the domain requirement and when to use hyperlink vs an http lib

headwinds:
First off - your library is fantastic - thank you for creating it!!!

I went through the Twitter auth today and had a couple hiccups:

1. Twitter requires http://127.0.0.1:5000 not http://localhost:5000 in the callback
2. I was attempting to use axios from the frontend instead of a simple hyperlink. I wanted to point this out in the quick start because I bet a lot of modern devs will probably try to use a library instead of a hyperlink if they haven't done the oath dance before. 

[demo](https://probe.now.sh/)

### DIFF
--- a/docs/quickstarts/twitter.rst
+++ b/docs/quickstarts/twitter.rst
@@ -60,7 +60,7 @@ immediately.
 In your view, make sure you use a plain hyperlink to begin the dance so that redirect loads Twitter's page in the browser. 
 
 ```
-<a href="http://127.0.0.1:5000/twitter/login">Sign in with Twitter</a>
+<a href="/twitter/login">Sign in with Twitter</a>
 ```
 
 You can only use http libraries like axios to check the authentication status. For instance, when the component mounts, you can call a function to check 
--- a/docs/quickstarts/twitter.rst
+++ b/docs/quickstarts/twitter.rst
@@ -5,7 +5,7 @@ Set up the application
 ----------------------
 Go to Twitter Application Manager at https://apps.twitter.com and create a
 new app. The application's "Callback URL" must be
-``http://127.0.0.1:5000/login/twitter/authorized``. `localhost:5000` will not work!
+``http://127.0.0.1:5000/login/twitter/authorized``. The domain ``localhost`` will not work with Twitter!
 Take note of the "API Key" and "API Secret" for the application.
 
 
@@ -54,48 +54,51 @@ run:
     $ export OAUTHLIB_INSECURE_TRANSPORT=1
     $ python twitter.py
 
-Visit http://127.0.0.1:5000 in your browser, and you should start the OAuth dance
-immediately.
+Visit ``http://127.0.0.1:5000`` in your browser, and you should start the OAuth dance
+immediately after you click on the Signin with Twitter hyperlink.
 
-In your view, make sure you use a plain hyperlink to begin the dance so that the redirect loads Twitter's confirmation page in the browser. 
+In your view, make sure you use a plain hyperlink to begin the dance
+so that the redirect loads Twitter's confirmation page in the browser.
 
-```
-<a href="/twitter/login">Sign in with Twitter</a>
-```
+.. code-block:: HTML
 
-You can only use http libraries like axios to check the authentication status. 
+    <a href="/twitter/login">Sign in with Twitter</a>
 
-For instance, when the component mounts, you can call a function to check the authentication status. 
+You can only use http libraries like axios to check the
+authentication status.
 
-If the authentication fails, then you can display a hyperlink for the user to begin the dance. 
+For instance, when the component mounts, you can call a function to check
+the authentication status. If the authentication fails, then
+you can display a hyperlink for the user to begin the dance, or
+if it succeeds, then simply display the authenticated username.
 
-If it succeeds, then simply display the authenticated username. 
+The following example uses axios and vuejs but this could be ported to
+react, angular, or vanila javascript. You could create 3 state variables:
+welcome (String ""), authenticated (Boolean false),
+authenticateCheckComplete (Boolean false), and then use these to
+either show the hyperlink or the authenticated username.
 
-```
-// an example using axios and vuejs but this could be ported to react, angular, or vanilajs
-// create 3 state variables: welcome (String ""), authenticated (Boolean false), authenticateCheckComplete (Boolean false)
-// if (!authenticated && authenticateCheckComplete) show the hyperlink 
-// if (authenticated && authenticateCheckComplete) show the username 
+.. code-block:: javascript
 
-const url = (document.domain === "127.0.0.1") ? 'http://127.0.0.1:5000/twitter/auth' : 'https://your-production-domain/twitter/auth'
+    function checkAuthentication(){
+        const self = this;
 
-function checkAuthentication(){
-    const self = this;
+        const url = (document.domain === "127.0.0.1")
+            ? 'http://127.0.0.1:5000/twitter/auth' : 'https://your-production-domain/twitter/auth'
 
-    axios.get(url).then( 
-        response => {
-            if (response.data.screen_name) {
-                self.welcome = "welcome " + response.data.screen_name;
-                self.authenticated = true;
+        axios.get(url).then(
+            response => {
+                if (response.data.screen_name) {
+                    self.welcome = "welcome " + response.data.screen_name;
+                    self.authenticated = true;
+                }
             }
-        }
-    ).catch(error => {
-        this.errored = error
-    }).finally(() => self.authenticateCheckComplete = true);
+        ).catch(error => {
+            this.errored = error
+        }).finally(() => self.authenticateCheckComplete = true);
 
-}
+    }
 
-```
 
 .. warning::
     :envvar:`OAUTHLIB_INSECURE_TRANSPORT` should only be used for local testing
--- a/docs/quickstarts/twitter.rst
+++ b/docs/quickstarts/twitter.rst
@@ -57,25 +57,27 @@ run:
 Visit http://127.0.0.1:5000 in your browser, and you should start the OAuth dance
 immediately.
 
-In your view, make sure you use a plain hyperlink to begin the dance so that redirect loads Twitter's page in the browser. 
+In your view, make sure you use a plain hyperlink to begin the dance so that the redirect loads Twitter's confirmation page in the browser. 
 
 ```
 <a href="/twitter/login">Sign in with Twitter</a>
 ```
 
-You can only use http libraries like axios to check the authentication status. For instance, when the component mounts, you can call a function to check 
-the authentication status. 
+You can only use http libraries like axios to check the authentication status. 
+
+For instance, when the component mounts, you can call a function to check the authentication status. 
 
 If the authentication fails, then you can display a hyperlink for the user to begin the dance. 
 
 If it succeeds, then simply display the authenticated username. 
 
 ```
+// an example using axios and vuejs but this could be ported to react, angular, or vanilajs
 // create 3 state variables: welcome (String ""), authenticated (Boolean false), authenticateCheckComplete (Boolean false)
-// !authenticated && authenticateCheckComplete show the hyperlink 
-// authenticated && authenticateCheckComplete show the username 
+// if (!authenticated && authenticateCheckComplete) show the hyperlink 
+// if (authenticated && authenticateCheckComplete) show the username 
 
-const url = (document.domain === "127.0.0.1") ? 'http://127.0.0.1:5000/twitter/auth' : 'https://production-domain/twitter/auth'
+const url = (document.domain === "127.0.0.1") ? 'http://127.0.0.1:5000/twitter/auth' : 'https://your-production-domain/twitter/auth'
 
 function checkAuthentication(){
     const self = this;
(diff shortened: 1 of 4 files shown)


docs/quickstarts/twitter.rst
```diff
@@ -54,9 +54,49 @@ run:
...
+            }
+        }
+    ).catch(error => {
+        this.errored = error
+    }).finally(() => self.authenticateCheckComplete = true);
+
+}
+
+```
+
```
daenney: This document is not Markdown, but ReStructuredText. As such how to do code blocks etc is quite different and this now renders rather broken documentation. Feel free to click on the "View File" button to see what I mean.

Please take a quick look at https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html and the rest of our documentation on how to do this correctly.

docs/quickstarts/twitter.rst
```diff
@@ -54,9 +54,49 @@ run:
     $ export OAUTHLIB_INSECURE_TRANSPORT=1
     $ python twitter.py
 
-Visit http://localhost:5000 in your browser, and you should start the OAuth dance
+Visit http://127.0.0.1:5000 in your browser, and you should start the OAuth dance
 immediately.
 
+In your view, make sure you use a plain hyperlink to begin the dance so that the redirect loads Twitter's confirmation page in the browser. 
+
+```
```
daenney: This is not how you do code blocks in rst files. They start with `.. code-block:: python` or whichever language as you can see in the rest of the file. The code then needs to be indented by 4 spaces.

docs/quickstarts/twitter.rst
```diff
@@ -54,9 +54,49 @@ run:
...
+
+You can only use http libraries like axios to check the authentication status. 
+
+For instance, when the component mounts, you can call a function to check the authentication status. 
+
+If the authentication fails, then you can display a hyperlink for the user to begin the dance. 
+
+If it succeeds, then simply display the authenticated username. 
+
+```
```
daenney: Idem.

docs/quickstarts/twitter.rst
headwinds: @daenney ahhh whoops sorry about that - obviously  I missed the rst format - so I've now research the format, added an rst plugin and linter to my vs code and have fixed those issues. 

thanks for the review!

docs/quickstarts/twitter.rst
headwinds: @daenney Ok when I click View File now - the file seems to be rendering correctly

daenney:
Awesome, thank you! I'll let @singingwolfboy have a look in case there's anything from their side since we usually only include Python code samples, but I think it does make sense in this case.

headwinds:
@daenney thank you - yeah I was a bit torn whether or not to include the frontend code but I believe it really shows how flask-dance integrates with flask, providing the complete flow and hopefully helps the next developer especially those new to oath

docs/quickstarts/twitter.rst
```diff
@@ -5,7 +5,7 @@ Set up the application
 ----------------------
 Go to Twitter Application Manager at https://apps.twitter.com and create a
 new app. The application's "Callback URL" must be
-``http://localhost:5000/login/twitter/authorized``.
+``http://127.0.0.1:5000/login/twitter/authorized``. The domain ``localhost`` will not work with Twitter!
```
singingwolfboy: Can you point me to some documentation for Twitter's OAuth implementation that supports this change?

docs/quickstarts/twitter.rst
```diff
@@ -24,7 +24,7 @@ Code
     )
     app.register_blueprint(blueprint, url_prefix="/login")
 
-    @app.route("/")
+    @app.route("/twitter/login")
```
singingwolfboy: Why are you changing this URL?

docs/quickstarts/twitter.rst
```diff
@@ -54,8 +54,51 @@ run:
     $ export OAUTHLIB_INSECURE_TRANSPORT=1
     $ python twitter.py
 
-Visit http://localhost:5000 in your browser, and you should start the OAuth dance
-immediately.
+Visit ``http://127.0.0.1:5000`` in your browser, and you should start the OAuth dance
+immediately after you click on the Signin with Twitter hyperlink.
```
singingwolfboy: Which hyperlink are you referring to? This is confusing.

docs/quickstarts/twitter.rst
```diff
@@ -54,8 +54,51 @@ run:
...
+
+In your view, make sure you use a plain hyperlink to begin the dance
+so that the redirect loads Twitter's confirmation page in the browser.
```
singingwolfboy: This sentence is full of jargon, and hard to understand if you are new to OAuth. Which view are you referring to? How is a "plain hyperlink" different from other options? By "dance", do you mean OAuth or something else? Which redirect are you referring to? What if I don't want to load Twitter's confirmation page in the browser?

docs/quickstarts/twitter.rst
```diff
@@ -54,8 +54,51 @@ run:
...
+
+.. code-block:: HTML
+
+    <a href="/twitter/login">Sign in with Twitter</a>
+
+You can only use http libraries like axios to check the
+authentication status.
```
singingwolfboy: What is axios, and why do I have to use it? What do you mean by "authentication status"? Are you claiming that it's impossible to check whether the user is logged in using Python, and Javascript is the only option?

docs/quickstarts/twitter.rst
```diff
@@ -54,8 +54,51 @@ run:
...
+
+For instance, when the component mounts, you can call a function to check
```
singingwolfboy: Why are we talking about components?

docs/quickstarts/twitter.rst
```diff
@@ -54,8 +54,51 @@ run:
...
+the authentication status. If the authentication fails, then
+you can display a hyperlink for the user to begin the dance, or
+if it succeeds, then simply display the authenticated username.
+
+The following example uses axios and vuejs but this could be ported to
+react, angular, or vanila javascript. You could create 3 state variables:
+welcome (String ""), authenticated (Boolean false),
+authenticateCheckComplete (Boolean false), and then use these to
+either show the hyperlink or the authenticated username.
```
singingwolfboy: Why are we talking about Javascript libraries in the Twitter quickstart? This doesn't belong here at all.

docs/quickstarts/twitter.rst
headwinds: @singingwolfboy I found it [here](https://twittercommunity.com/t/why-cant-i-use-localhost-as-my-oauth-callback/708) - I also assumed localhost would work until I tried it - in the app callback settings, it even accepts localhost without a validation error but it fails when I tried it use it.

docs/quickstarts/twitter.rst
headwinds: @singingwolfboy sorry I was confused here - I'll change it back - I thought one would need a dedicated route for twitter auth and that the "/" route would be used to return your index page.

docs/quickstarts/twitter.rst
headwinds: Because Twitter wants 127.0.0.1 instead of localhost, I thought it should be consistent.

docs/quickstarts/twitter.rst
headwinds: yes I'll re-write this and attempt to clean up the jargon - good point about possibly not wanting to load the confirmation page in the browser - I didn't consider that.

docs/quickstarts/twitter.rst
headwinds: good point - I'm thinking too much from the frontend perspective - I'll re-write this section in the new doc.

docs/quickstarts/twitter.rst
headwinds: Most frontends today involve components (React, Angular, Vuejs are all component based) but yes this should be moved into another doc directed at a modern frontend audience.

docs/quickstarts/twitter.rst
headwinds: As you recommended, I'll move this into a new doc directed at React/Angular/VueJS devs and then we can link to the new document. That's a great compromise - thanks!

headwinds:
@daenney I'll work on this change now and it should be ready by monday:
- create a new file under the docs directory to write some documentation about using JS frameworks with Flask-Dance; work in the the feedback and re-write some sections for clarity
- add a link to this file from the bottom of the Twitter quickstart

headwinds:
@daenney I've completed that update so there is a new twitter-frontend doc and the twitter doc has been reverted and simply has a link to it at the end. I also created a new repo [flask-dance-twitter-frontend](https://github.com/headwinds/flask-dance-twitter-frontend) as a demo. 

I also attempted to squash my last 8 commits into 1 but for some reason they are still showing up above but my git log reflects only 2 commits; not sure why that is?!

singingwolfboy:
I really appreciate your efforts so far @headwinds, but I think we should take a step back and think about the goal of this documentation. When someone is reading developer documentation, they typically have certain questions that they want answered, and they are looking for the quickest, simplest, most direct answers to those questions -- so, the goal should be to answer those questions in the quickest, simplest, most direct way possible, without over-simplifying or misleading the reader.

Here are some questions that I can imagine a reader might have, regarding Flask-Dance and front-end development:

* Is it possible to use Flask-Dance for front-end development?
* Does it only work for certain OAuth providers, or for all OAuth providers? Does it work for the specific provider that I want to use?
* Does it only work for certain JS libraries/frameworks, or for all JS libraries/frameworks? Does it work for the specific library/framework that I want to use?
* What is the simplest possible code I can use as a proof-of-concept, so I can verify for myself that it works?
* How do the backend and the frontend work together? Do I need to modify my backend to work with my frontend, or vice-versa?
* What are the benefits and limitations of doing this on the frontend, rather than the backend? If I have a choice, when should I choose the frontend, and when should I choose the backend?
* Are there any "gotchas", or unexpected problems, that I might run into?
* If I'm focused on doing OAuth in the frontend, what do I even need a backend for? What is the benefit of using Flask-Dance in this approach?

There are probably many, many more questions that I'm not thinking of -- these are just the ones that immediately come to mind.

The document that you've written so far seems to make the following points:

* Check out this example repository that uses Flask-Dance with Twitter
* Here's how to run this example repository
* Here's a snippet of code that calls a custom API from this repository

Do you see how these points do not match with the questions above? In fact, they will probably lead to *more* confusion and frustration. For example:

* Does this mean that you can *only* do frontend OAuth with Twitter and axios?
* How do I learn from this repository? There's a lot of code here, and most of it is not relevant to my questions...
* I don't actually care about the way *you* chose to do it, I want to do it *my* way, so your example isn't helpful...

@headwinds, I want to emphasize that I really appreciate you making this pull request, and bringing up this topic in the first place. Some documentation for how to use Flask-Dance on the frontend would be really helpful for a lot of people! But unfortunately, I don't think the doc you've put together is going to reduce confusion, so I'm not going to pull it in. If you'd like to try rewriting it again, maybe using the questions I listed above as a guide, then I'd be happy to take another look at it if you want!

headwinds:
@daenney Yes I see - I appreciate your feedback - seems like you've created an excellent skeleton for the right documentation and probably makes more sense that should answer them. 

You're right - I only wanted to provide some sample code and how I happened to solve it which I thought might help a few others in a similar situation.  I really just provided the shallow documentation that I was looking for so I don't mind maintaining it my own repo. If anyone has further questions, I'll definitely promote your repo for a deeper dive. I realize now that I've simply made the common open source mistake of not pitching the idea before my PR. Live & learn ;-D

singingwolfboy:
Live & learn, indeed! Thanks for understanding the situation.

One other thing I'll mention: I'm not at all opposed to including a list of open source projects that *use* Flask-Dance, so that someone browsing the documentation could find those links and read the code, if they're interested. I'm imagining something where each project has a title, a link to the project, and a (very) short description of what it is and how it uses Flask-Dance. Maybe a bit like the [Flask extensions registry](http://flask.pocoo.org/extensions/). Would you like to create that page, and include your project in it?

If someone else finds that documentation page and clicks on your project, then they'll probably have lots of other questions that are related *specifically* to your project: questions like "What does this project do?" and "How do I run it locally?". So I definitely recommend that you maintain your *own* set of documentation with your project, which could be as simple as a README file if you want.

Does this seem like a good compromise?

headwinds:
Yes that does seem fair - in fact, I know there are several github example repos demonstrating flask-dance against different social apis like this [one also for twitter but in Spanish](https://github.com/jeaanmichel/flask-dance-twitter-login-example).

How about I create rst doc entitled "Flask-dance Registry" or "Projects using Flask-dance? 

An rst doc something like this:

```
Projects using Flask-dance

Contributing: please create a branch and make a PR to add your project. Ideally, each project should contain a readme, source code and hosted demo. 

`flask-dance-twitter-login-example <https://github.com/jeaanmichel/flask-dance-twitter-login-example>`_
description: Um simples app para fazer login no Twitter
language: Spanish  
auth: Twitter 

`flask-dance-twitter-frontend <https://github.com/jeaanmichel/flask-dance-frontend>`_
description: an example project that has VueJS as the frontend and Flask Python as the backend
language: English   
auth: Twitter 

`github-heroku-attack-rabbits <https://github.com/charlesreid1/github-heroku-attack-rabbits>`_
description: running a Python Flask server uses Flask-Dance to authenticate visitors with Github which allows you fine-grained access control 
language: English   
auth: Github 

```

Would that work for you?

PR #208, opened 2019-01-13
Add a list of projects that use Flask-Dance

singingwolfboy:
Follow up to #202. @headwinds, how does this look to you?


headwinds:
@singingwolfboy looks good! ha obviously I forgot to make a PR for the [version I did](https://github.com/headwinds/flask-dance/blob/master/docs/quickstarts/registry.rst) but your version does the trick too

PR #212, opened 2019-01-20
Fix a warning in setup.py

felixonmars:
```
Warning: 'classifiers' should be a list, got type 'tuple'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
     },
     cmdclass = {'test': PyTest},
     license='MIT',
-    classifiers=(
+    classifiers=[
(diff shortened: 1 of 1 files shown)


singingwolfboy:
How did you get this warning?

felixonmars:
@singingwolfboy Just run "python setup.py" with or without any additional command. This may be coming from Python 3.7.

felixonmars:
Update: found it, so it's indeed coming from Python 3.7: https://bugs.python.org/issue19610

daenney:
I really don't understand why they decided that a tuple wasn't good enough now but :man_shrugging:.

PR #217, opened 2019-01-27
add parameter include_client_id.

### LINKED ISSUE #218: `include_client_id` in requests oauthlib 1.2.0 is not clear
Migrating from `requests-oauthlib` 1.0.0 to 1.2.0, Instagram stopped working.
After some debugging, I found out that the `client_id` and `client_secret` are not being sent. Long story short, I've added `token_url_params={'include_client_id': True}` to the initiation of the Instagram blueprint.
I think it should be more clear how to use the `include_client_id` parameter, and pass it as a standalone param instead of inside a dict.

### LINKED ISSUE #363: Flask 2.0
Pallets just released 2.0: https://flask.palletsprojects.com/en/2.0.x/changes/#version-2-0-0. I'm just dropping this issue here as a reminder for us to review the changelog and see if there's anything we need to do.

AdamGold:
closes #218

### DIFF
--- a/flask_dance/consumer/oauth2.py
+++ b/flask_dance/consumer/oauth2.py
@@ -38,6 +38,7 @@ class OAuth2ConsumerBlueprint(BaseOAuthConsumerBlueprint):
             authorization_url_params=None,
             token_url=None,
             token_url_params=None,
+            include_client_id=False,
             redirect_url=None,
             redirect_to=None,
             session_class=None,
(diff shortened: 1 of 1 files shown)


flask_dance/consumer/oauth2.py
```diff
@@ -243,6 +247,7 @@ def authorized(self):
                 self.token_url,
                 authorization_response=request.url,
                 client_secret=self.client_secret,
+                include_client_id=self.include_client_id,
```
singingwolfboy: Why not just set this in `token_url_params`?

AdamGold: Since `oauth2_session.py` lists this as a parameter, I think it would be more clear if this was a standalone parameter in this method as well.

singingwolfboy: Most people using Flask-Dance shouldn't have to think about requests-oauthlib, or whether it's correct to include the client ID, or any of those fiddly OAuth details. The whole point of Flask-Dance is to make it simpler. `token_url_params` provides an opportunity for developers who know those fiddly details to configure them as necessary, but including this option in the top-level arguments (and adding documentation for that option) seems needlessly confusing to me.

PR #220, opened 2019-01-30
add before_login signal to allow custom actions before first redirect

### LINKED ISSUE #219: Add option to insert code before `login` redirect
Found myself wanting to create a session before the `login` redirect. For example, A user clicks "Login with Facebook" -> getting redirected to `http://example.com/facebook` -> CREATE SESSION -> redirect to Facebook page -> `http://example.com/facebook/authorized?....`.

AdamGold:
Implement `before_authorized` signal to allow custom methods to run before the initial network redirection (in the `login` method). closes #219.

### DIFF
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -64,5 +64,23 @@ The following signals exist in Flask-Dance:
     anyway, so it is your responsibility to hook into this signal and inform
     the user that there was an error.
 
+
+.. data:: oauth_before_login
+
+    This signal is sent before redirecting to the provider login page. The signal
+    is sent with a ``url`` parameter specifying the redirect URL. This signal is mostly
+    useful for doing things like session construction / de-construction before the user
+    is redirected.
+
+    Example subscriber::
+
+        from flask_dance.consumer import oauth_before_login
+
+        @oauth_before_login.connect
+        def before_login(blueprint, url):
+            flask.session["i_was_here_before"] = "the_login"
+            # if "facebook" in url:
+                # do facebook stuff
+
 .. _flash a message: http://flask.pocoo.org/docs/latest/patterns/flashing/
 .. _blinker: http://pythonhosted.org/blinker/
--- a/flask_dance/__init__.py
+++ b/flask_dance/__init__.py
@@ -3,4 +3,4 @@ from __future__ import unicode_literals
 
 from .consumer import OAuth1ConsumerBlueprint, OAuth2ConsumerBlueprint
 
-__version__ = "1.3.0"
+__version__ = "1.3.1"
--- a/flask_dance/__init__.py
+++ b/flask_dance/__init__.py
@@ -3,4 +3,4 @@ from __future__ import unicode_literals
 
 from .consumer import OAuth1ConsumerBlueprint, OAuth2ConsumerBlueprint
 
-__version__ = "1.3.1"
+__version__ = "1.3.0"
--- a/flask_dance/consumer/base.py
+++ b/flask_dance/consumer/base.py
@@ -13,7 +13,7 @@ from flask_dance.utils import getattrd, timestamp_from_datetime
 
 _signals = Namespace()
 oauth_authorized = _signals.signal('oauth-authorized')
-oauth_before_login = _signals.signal('oauth-before-authorized')
+oauth_before_login = _signals.signal('oauth-before-login')
 oauth_error = _signals.signal('oauth-error')
 
 
--- a/flask_dance/consumer/base.py
+++ b/flask_dance/consumer/base.py
@@ -13,7 +13,7 @@ from flask_dance.utils import getattrd, timestamp_from_datetime
 
 _signals = Namespace()
 oauth_authorized = _signals.signal('oauth-authorized')
-oauth_before_authorized = _signals.signal('oauth-before-authorized')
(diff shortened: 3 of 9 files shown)


singingwolfboy:
I like this contribution! Thank you for adding tests. 😄 

However, why not write a pre-login view that would do what you need? For example:

```python
@app.route("/")
def index():
    if not github.authoirzed:
        return redirect(url_for("prelogin"))
    # implement the rest of your view here

@app.route("/prelogin")
def prelogin():
    # do whatever you need to do before login begins
    return redirect("github.login")
```

Does this suit your use-case? I'm not opposed to adding another signal, but I just want to be sure that you've considered the other options first.

If we *do* want to go for the signal, I suggest naming it `before_login`, since that is more consistent with the terminology of this library. In addition, we'll need some kind of documentation to let people know that this signal exists: you could add it to the `docs/signals.rst` file. We also need to add something to the `CHANGELOG.rst` file.

AdamGold:
It's definitely an option, but it would mean I'd have to implement a new route for each of the networks. Why not have one method to rule them all, considering I want to do the same action in all of them?

About the renaming and doc - will do.

singingwolfboy:
Perfectly valid reason. 👍 Let's implement this signal!

AdamGold:
All done. Updated to ``1.3.1`` in CHANGELOG, let me know if that's okay.

flask_dance/__init__.py
```diff
@@ -3,4 +3,4 @@
 
 from .consumer import OAuth1ConsumerBlueprint, OAuth2ConsumerBlueprint
 
-__version__ = "1.3.0"
+__version__ = "1.3.1"
```
singingwolfboy: Please don’t touch the version number. Put the changelog entries in the “unreleased” section.

docs/signals.rst
```diff
@@ -64,5 +64,23 @@ The following signals exist in Flask-Dance:
...
+.. data:: oauth_before_login
+
+    This signal is sent before redirecting to the provider login page. The signal
+    is sent with a ``url`` parameter specifying the redirect URL. This signal is mostly
+    useful for doing things like session construction / de-construction before the user
+    is redirected.
+
+    Example subscriber::
+
+        from flask import flash
```
singingwolfboy: Don’t import things you don’t use

docs/signals.rst
```diff
@@ -64,5 +64,23 @@ The following signals exist in Flask-Dance:
...
+        from flask_dance.consumer import oauth_before_login
+
+        @oauth_before_login.connect
+        def before_login(blueprint, url):
+            flask.session["i_was_here_before"] = "the_login"
+            // do something with URL
```
singingwolfboy: What sort of things could you do with the URL here? Can you provide an example?

flask_dance/consumer/base.py
```diff
@@ -13,6 +13,7 @@
 
 _signals = Namespace()
 oauth_authorized = _signals.signal('oauth-authorized')
+oauth_before_login = _signals.signal('oauth-before-authorized')
```
singingwolfboy: Names should match

tests/consumer/test_oauth1.py
```diff
@@ -323,6 +323,22 @@ def callback(*args, **kwargs):
     assert len(calls) == 1
 
 
+@requires_blinker
+def test_signal_oauth_before_login(request):
+    app, bp = make_app()
+    bp.session.fetch_request_token = mock.Mock(return_value="test-token")
+    def callback(*args, **kwargs):
+        flask.session["test"] = "test"
```
singingwolfboy: Nitpick: can you make this test do something a little more like a real-world use-case?

tests/consumer/test_oauth2.py
```diff
@@ -453,6 +453,20 @@ def callback(*args, **kwargs):
     assert calls[0][1] == {"token": fake_token}
 
 
+@requires_blinker
+def test_signal_oauth_before_login(request):
+    app, bp = make_app()
+    def callback(*args, **kwargs):
+        flask.session["test"] = "test"
```
singingwolfboy: Same here

tests/consumer/test_oauth1.py
AdamGold: I use it to log the user out before the login. Here's why:
I am using Flask as an API service, so the flow is:
User clicks on "login with X (facebook, instagram etc.)" from mobile --> browser opens with the server. Now here's the gotcha - I need to know if the user is authenticated - so the server can decide whether it needs to associate a new network for a specific user, or create a new user. I need a login session opened and closed for each request, but before opening a new session - I want to clear any unwanted ones that may exist - and I do it with `before_login`.

AdamGold: What I'm trying to say is that I think that saving a session before the redirect is indeed a real world use case.

singingwolfboy: Sure. Then can you do something like “flask.session[“google_id”] = 123” before the code being tested, then do “del flask.session[“google_id”]” in the signal handler, then assert that there is no “google_id” in the Flask session at the end of the test?

AdamGold: Done.

singingwolfboy:
Thanks a lot! 🎉

PR #222, opened 2019-02-03
Set up code formatting with black

singingwolfboy:
https://github.com/ambv/black


daenney:
:100: 

Seems like we need to let Black reformat some stuff before this can merge.

PR #227, opened 2019-02-21
Google OAuth prompt update

jmprovencher:
Adding the possibility to reprompt the select account page with Google OAuth.

Please see the new documentation for 'prompt' [here](https://developers.google.com/identity/protocols/OpenIDConnect#authenticationuriparameters)

resolves #227

### DIFF
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -48,7 +48,6 @@ def make_google_blueprint(
             consent. Defaults to False
         reprompt_select_account (bool): If True, force Google to re-prompt the select account page,
          even if there is a single logged-in user. Defaults to False
-            consent. Defaults to False
         redirect_url (str): the URL to redirect to after the authentication
             dance is complete
         redirect_to (str): if ``redirect_url`` is not defined, the name of the
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -47,7 +47,7 @@ def make_google_blueprint(
             for their consent, even if the user has already given their
             consent. Defaults to False
         reprompt_select_account (bool): If True, force Google to re-prompt the select account page,
-         even if there is a single logged-in user. Defaults to False
+            even if there is a single logged-in user. Defaults to False
         redirect_url (str): the URL to redirect to after the authentication
             dance is complete
         redirect_to (str): if ``redirect_url`` is not defined, the name of the
@@ -116,7 +116,6 @@ def make_google_blueprint(
     if prompt_params:
         prompt_params = " ".join(prompt_params)
         authorization_url_params["prompt"] = prompt_params
-
     if hosted_domain:
         authorization_url_params["hd"] = hosted_domain
     google_bp = OAuth2ConsumerBlueprint(
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -19,6 +19,7 @@ def make_google_blueprint(
     scope=None,
     offline=False,
     reprompt_consent=False,
+    reprompt_select_account=False,
     redirect_url=None,
     redirect_to=None,
     login_url=None,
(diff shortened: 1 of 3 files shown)


flask_dance/contrib/google.py
```diff
@@ -45,6 +46,9 @@ def make_google_blueprint(
         reprompt_consent (bool): If True, force Google to re-prompt the user
             for their consent, even if the user has already given their
             consent. Defaults to False
+        reprompt_select_account (bool): If True, force Google to re-prompt the select account page,
+         even if there is a single logged-in user. Defaults to False
+            consent. Defaults to False
```
daenney: Too much copy paste?

jmprovencher: I tried to keep the same syntax. Can reformulate if you want?

jmprovencher: Good catch.

jmprovencher:
What is the flow with the contribution process? When this PR is merged when can I expect to be able to download the latest version through Pip?

flask_dance/contrib/google.py
```diff
@@ -45,6 +46,8 @@ def make_google_blueprint(
         reprompt_consent (bool): If True, force Google to re-prompt the user
             for their consent, even if the user has already given their
             consent. Defaults to False
+        reprompt_select_account (bool): If True, force Google to re-prompt the select account page,
+         even if there is a single logged-in user. Defaults to False
```
singingwolfboy: Can you please make the indentation here match the indentation for the other parameter docstrings?

flask_dance/contrib/google.py
```diff
@@ -101,12 +104,19 @@ def logged_in(blueprint, token):
...
         auto_refresh_url = "https://accounts.google.com/o/oauth2/token"
     if reprompt_consent:
-        authorization_url_params["approval_prompt"] = "force"
+        prompt_params.append("consent")
+    if reprompt_select_account:
+        prompt_params.append("select_account")
+    if prompt_params:
+        prompt_params = " ".join(prompt_params)
+        authorization_url_params["prompt"] = prompt_params
+
```
singingwolfboy: Please remove this blank line. (Or is it there intentionally?)

singingwolfboy:
Oh also, can you add an entry to the CHANGELOG?

jmprovencher:
> > When this PR is merged when can I expect to be able to download the latest version through Pip?
> 
> I typically try to wait until several PRs have merged before cutting a new release, so that there isn't as much churn. However, several PRs have already merged (including #225, which is a big one), so I'll probably cut and publish Flask-Dance 1.4.0 shortly after this merges.
> 
> > What is the flow with the contribution process?
> 
> I'm not quite sure what you're asking... can you clarify your question a bit?

@singingwolfboy 
Simply because I am using Flask-Dance on a production project and I would like to use this feature to fix a behaviour on my end :).

I will make the requested changes.

jmprovencher:
@singingwolfboy I made the requested changes and I added pytest fixture for google blueprint.

tests/contrib/test_google.py
```diff
@@ -109,57 +109,64 @@ def test_context_local():
...
+        app.secret_key = "backups"
+        goog_bp = make_google_blueprint("foo", "bar", **kwargs)
+        app.register_blueprint(goog_bp)
+
+        with app.test_client() as client:
+            return client.get(
+                "/google",
+                base_url="https://a.b.c",
+                follow_redirects=False,
+            )
```
daenney: I don't like the idea of this being in a fixture. If we now want to do something else we can't access the `app` instance at all. I'd much rather this returns `app` instead of the response from `client.get`.

singingwolfboy: I agree with @daenney. Since this seems like it's going to be a non-trivial amount of work to refactor these tests, could you just put them back to the repetitive state that they were in before? We can refactor them in a different pull request.

jmprovencher: I could have refactor this and return the app object. For the moment I reverted and put them back in their repetitive state.

singingwolfboy:
FYI, I played around with reducing the repetition in the tests, and came up with this: https://github.com/singingwolfboy/flask-dance/commit/40ac51253999e8c4c79e0531a17a98767e581367

It's a larger change than I expected, and doesn't actually make the tests _much_ shorter, but it eliminates some needless repetition without making the tests more difficult to read. I'm glad I decided to separate that refactor out of this PR!

singingwolfboy:
@jmprovencher Flask-Dance 1.4.0 is published on PyPI, with this change included!

jmprovencher:
Nice refactoring of the tests! Thanks for the quick release ! 😲

jmprovencher:
@singingwolfboy If you ever need contributor to review feel free to ask!

PR #229, opened 2019-02-22
Consistent config names

singingwolfboy:
Previously, I tried to make the pre-set configurations pull values from the Flask config based on the terminology used on the provider website. For example, Twitter calls the credentials "API Key" and "API Secret", even though the OAuth standard calls them "client key" and "client secret". This resulted in inconsistency across providers.

This pull request makes the pre-set configurations consistent with the OAuth standard.

This is a **backwards-incompatible** change. As such, it should only be merged just before we release Flask-Dance 2.0.

### DIFF
--- a/tests/contrib/test_zoho.py
+++ b/tests/contrib/test_zoho.py
@@ -43,18 +43,6 @@ def test_load_from_config(make_app):
 
 
 @responses.activate
-def test_load_from_params(make_app):
-    app = make_app(client_id="not_foo", client_secret="not_bar")
-    app.config["ZOHO_OAUTH_CLIENT_ID"] = "foo"
-    app.config["ZOHO_OAUTH_CLIENT_SECRET"] = "bar"
-
-    resp = app.test_client().get("/zoho")
-    url = resp.headers["Location"]
-    client_id = URLObject(url).query.dict.get("client_id")
-    assert client_id == "not_foo"
-
-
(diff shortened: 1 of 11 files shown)


tests/contrib/test_zoho.py
```diff
@@ -42,18 +42,6 @@ def test_load_from_config(make_app):
...
-def test_load_from_params(make_app):
-    app = make_app(client_id="not_foo", client_secret="not_bar")
-    app.config["ZOHO_OAUTH_CLIENT_ID"] = "foo"
-    app.config["ZOHO_OAUTH_CLIENT_SECRET"] = "bar"
-
-    resp = app.test_client().get("/zoho")
-    url = resp.headers["Location"]
-    client_id = URLObject(url).query.dict.get("client_id")
-    assert client_id == "not_foo"
-
```
daenney: Did you intend to remove this test?

singingwolfboy: Yes, I did. It was asserting that the params passed in the `create_zoho_blueprint()` function should override the values in the app config. This does not match the behavior with any of the other blueprints, so I changed the behavior and deleted the test.

Do you think it was the right thing to do? I'm curious to hear your opinion.

daenney: Makes sense. I always value consistency over magical unique feature.

PR #221, opened 2019-02-01
Reorganize docs

singingwolfboy:
The docs for this project need some love. This is an attempt to rewrite and reorganize them to make them clearer and more approachable.

# How do I see how it looks?
1. Clone this repository, and switch to the `reorganize-docs` branch
2. `pip install -r requirements.txt -r dev-requirements.txt`
3. `cd docs`
4. `make html`
5. Open `_build/html/index.html` in your browser

# Major changes
* Removed all the provider-specific quickstart files, in favor of linking to provider-specific example projects such as [flask-dance-github](https://github.com/singingwolfboy/flask-dance-github) and [flask-dance-google](https://github.com/singingwolfboy/flask-dance-google)
* Renamed "backend" to "storage" (see #223)
* Rewrote testing documentation to reference `MemoryBackend` (which is what we use in the tests for this project) and [Betamax](https://betamax.readthedocs.io)
* Added the following docs: "concepts", "understanding the magic", and "contributing"
* Reorganized docs into "User Guide", "Options & Configuration", and "Advanced Topics" sections

I would love some help with this! Any thoughts, suggestions, or feedback would be greatly appreciated.

PR #230, opened 2019-02-22
Remove `next` redirect

singingwolfboy:
Previously, Flask-Dance had an undocumented feature where it would automatically redirect based on a `next` parameter in the URL. This undocumented feature has been removed.

This is a **backwards incompatible** change. As such, it should only be merged just before we release Flask-Dance 2.0.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,9 @@ Changelog
 `unreleased`_
 -------------
 
-nothing yet
+* **Backwards incompatible:** Previously, Flask-Dance had an undocumented
+  feature where it would automatically redirect based on a ``next``
+  parameter in the URL. This undocumented feature has been removed.
 
 `1.4.0`_ (2019-02-22)
 ---------------------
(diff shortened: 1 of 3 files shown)


jmprovencher:
Why are you removing this feature out of curiosity? I was using it to redirect my user to my frontend.

daenney:
There's a documented feature on how to do this. This is the magical undocumented one that gets removed.

jmprovencher:
@daenney Do you have a link to the documented feature about this ?

singingwolfboy:
@jmprovencher [Here's the documentation for how to do dynamic redirects.](https://flask-dance.readthedocs.io/en/latest/understanding-the-magic.html#finishing-the-dance) To be fair, [I just wrote this documentation recently](https://github.com/singingwolfboy/flask-dance/pull/221), so it's understandable that you haven't seen it yet.

I should also point out that I would _love_ more assistance with the documentation for this project. If there is something that you find confusing or hard to find, please let me know -- or even better, send a pull request to the [docs directory](https://github.com/singingwolfboy/flask-dance/tree/master/docs)!

jmprovencher:
@singingwolfboy Superb. This piece of documentation wasn't there when I implemented Flask-Dance in my solution. I'll make sure to update my project with the latest practices. Thanks a lot.

daenney:
@singingwolfboy Should we start merging the "breaking changes" PRs are do you have changes you'd still like to do for an additional 1.x point release?

singingwolfboy:
@daenney I just created #231 to start planning this process. Let's continue the discussion there.

PR #234, opened 2019-02-28
Update logout.rst

doekman:
The stored token will never be used again so better delete it.
And also when the token is revoked `google.authorized` is still true, and that's unwanted. Deleting the token will make `google.authorized` false.

(at least that's my understanding. Correct me if I'm wrong)

### DIFF
--- a/docs/logout.rst
+++ b/docs/logout.rst
@@ -119,8 +119,6 @@ Here's some sample code that works with Google:
             headers={"Content-Type": "application/x-www-form-urlencoded"}
         )
         assert resp.ok, resp.text
-        # delete token from backend
-        blueprint.backend.delete(blueprint)
         logout_user()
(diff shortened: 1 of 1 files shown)


daenney:
I think you might need to do that the other way around. `logout()` could potentially be implemented in such a way that it also revokes the token, so you still need it at that point. After `logout()` it should always be safe to toss it.

doekman:
So order doesn't matter? I don't use flask_login myself. Without the logout_user() it works fine in my own project.

daenney:
I think it's fine either way. I got confused due to multiple things being named logout. I'll let @singingwolfboy decide.

singingwolfboy:
Fixed in https://github.com/singingwolfboy/flask-dance/commit/bcae95fb6301caf34515c163ebe10c81e5044397

Issue #231, opened 2019-02-26
Flask-Dance 2.0


singingwolfboy:
We're approaching the 2.0 release for Flask-Dance. Incrementing the major version is an opportunity to make backwards-incompatible changes. Here are some proposed changes we could make:

* drop `next` redirect (PR #230)
* consistent config names for pre-set configurations (PR #229)
* multithreading support? (issue #205)
* auto-refreshing tokens? (issue #143)
* remove special treatment of `expires_at` parameter? (https://github.com/oauthlib/oauthlib/issues/568)

We certainly don't have to make _all_ of these changes in version 2.0. In particular, while I would _love_ to get multithreading working, I don't even know how to properly test the code to verify it.

If anyone else has ideas of what to include for version 2.0, please add a comment to this issue.

daenney:
I think #205 will risk delaying 2.0 a lot. It seems to require changes in the supporting libraries we use, like lazy. We'd need to vet any other supporting library too to ensure it's thread-safe and do a lot of testing on our side to reveal any potential threading bugs. People generally don't think about thread safety when writing Python so this might prove to be a pretty big task requiring many changes in both our code and our dependencies.

There's all kinds of nastiness to keep in mind too, for example `counter = 1` is thread safe, but `counter = counter + 1` is not. CPython guarantees a lot of actions on built-in types and single operations to be thread safe, but that is an implementation detail of CPython and not necessarily true for PyPy etc.

Given how bad it could be if we were to leak things like tokens across threads we should be really careful about doing this.

singingwolfboy:
> I would much prefer not supporting this and recommending people run this in a multiprocess/prefork mode instead, which is what normally happens when you run with mod_wsgi or uwsgi.

Totally fair. Any chance I could ask you to put this recommendation into our docs? 😄

daenney:
Sure, will do!

daenney:
I just realised something... Flask itself remains thread safe due to its use of context locals and the app and request contexts. So in theory, we should be too as long as nothing is done outside of a context (aside from loading settings). I'm just not sure how this interacts with the import machinery and any supporting library but as long as we create the Oauth client within at least the app context we shouldn't risk leaking anything across threads.

singingwolfboy:
Discussion about how to fix multi-threading probably belongs in #205, not here 😄

daenney:
Good point :laughing:. I've replied there. Lets see what happens.

PR #238, opened 2019-03-15
Replace lazy dependency with werkzeug.utils.cached_property

singingwolfboy:
I didn't know about [`werkzeug.utils.cached_property`](http://werkzeug.pocoo.org/docs/0.14/utils/#werkzeug.utils.cached_property) until @daenney [mentioned it](https://github.com/singingwolfboy/flask-dance/issues/231#issuecomment-467408039). Looks like it does _almost_ what we need, except that it's missing the ability to invalidate the cache. [So I added a pull request to do just that.](https://github.com/pallets/werkzeug/pull/1474)

*Maybe* this makes Flask-Dance thread-safe? Could someone please check that?

### DIFF
--- a/flask_dance/consumer/base.py
+++ b/flask_dance/consumer/base.py
@@ -3,13 +3,20 @@ from __future__ import unicode_literals, print_function
 import warnings
 from datetime import datetime, timedelta
 import six
-from lazy import lazy
 from abc import ABCMeta, abstractmethod, abstractproperty
 from werkzeug.datastructures import CallbackDict
 import flask
 from flask.signals import Namespace
 from flask_dance.consumer.storage.session import SessionStorage
 from flask_dance.utils import getattrd, timestamp_from_datetime
+
+try:
+    from werkzeug.utils import invalidate_cached_property
+except ImportError:
+    from werkzeug._internal import _missing
+
+    def invalidate_cached_property(obj, name):
+        obj.__dict__[name] = _missing
 
 
 _signals = Namespace()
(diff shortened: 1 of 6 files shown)


daenney:
It at least solves the issue with the `lazy` library. We still need to review the rest of our code and figure out what would happen if our code were to use thread-unsafe dependency. I'm mostly worried about requests and oauthlib there as I don't think either is thread safe (perhaps they are, but afaik they weren't designed with that in mind so we might be in for some surprises).

flask_dance/consumer/base.py
```diff
@@ -3,14 +3,21 @@
...
 from flask_dance.utils import getattrd, timestamp_from_datetime
 
+try:
+    from werkzeug.utils import invalidate_cached_property
+except ImportError:
+    from werkzeug._internal import _missing
+
+    def invalidate_cached_property(obj, name):
+        obj.__dict__[name] = _missing
+
```
daenney: I don't quite understand this block, what is this magical `_missing` thing?

singingwolfboy: It's an implementation detail of how the `cached_property` decorator works. [If you check Werkzeug's source code](https://github.com/pallets/werkzeug/blob/master/src/werkzeug/utils.py#L89), you can see that `_missing` is a sentinel value that is returned if there is no value in `obj.__dict__[name]`, which works as a cache. In order to invalidate the cache, you have to reset that value to `_missing`, so that `cached_property` thinks that the cache is empty.

We should probably wait to merge this pull request until https://github.com/pallets/werkzeug/pull/1474 is merged -- at this point, the PR is still open, so the API could change.

daenney: Thanks, that explains a few things!

PR #244, opened 2019-03-21
Fix Werkzeug ProxyFix issue

singingwolfboy:
Due to Werkzeug 0.15

### DIFF
--- a/tests/consumer/test_oauth1.py
+++ b/tests/consumer/test_oauth1.py
@@ -9,7 +9,7 @@ import pytest
 import mock
 import responses
 import flask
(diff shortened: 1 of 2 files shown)


daenney:
Does this mean we need to set a minimum version for Werkzeug? Or does this still work with earlier versions?

singingwolfboy:
Notice that the only changes I made were in the tests, not the code being tested. The tests were broken, but not the code itself. Does that make sense?

To answer your question: no, we don't need to set a minimum version for Werkzeug.

daenney:
Good point! Thanks

PR #247, opened 2019-03-30
Backwards-incompatible changes for 2.0

### LINKED ISSUE #231: Flask-Dance 2.0

singingwolfboy:
Includes the following changes: #246, #238, #230, #229, #245. Closes #231.


daenney:
One concern I have is the amount of duplication we have with the `from werkzeug ... import cached_property` etc. Can we put that in a separate package, like `utils.py` and import `invalidate_cached_property` through there so we can put that logic in one place? Once we bump the requirement to werkzeug >= 1.0.0 we could then drop the import through utils.

singingwolfboy:
@daenney Good thought! Fixed.

PR #252, opened 2019-04-29
s/consumer/provider/

eblume:
Please let me know if this is incorrect. I'm just learning flask-dance but from what I know about oauth2, this was flipped.


daenney:
You're absolutely correct. Thanks for the fix!

PR #257, opened 2019-05-06
Will add blinker to list of requirements

alevis:
Blinker is used in `.travis.yml` oauth tests and in `setup.py`;
 - `extras_require={"sqla": ["sqlalchemy", "sqlalchemy-utils"], "signals": ["blinker"]},`

However, blinker is not installed since it's not part of `requirements.txt` which leads to `AttributeError: '_FakeSignal' object has no attribute 'connect_via'` when using google for oauth.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+blinker>=1.4
 requests>=2.0
 oauthlib
 requests-oauthlib>=1.0.0


daenney:
Blinkers is defined as an extra-requires: https://github.com/singingwolfboy/flask-dance/blob/master/setup.py#L64. You can install it with `pip install flask-dance[blinkers]`.

alevis:
> Blinkers is defined as an extra-requires, since it's not a core requirement of Flask-Dance: https://github.com/singingwolfboy/flask-dance/blob/master/setup.py#L64. You can install it with `pip install flask-dance[signals]`. You can specify it as such in your own `requirements.txt` too.

My point was `flask-dance` doesn't work right off the bat or when using the [example](https://flask-dance.readthedocs.io/en/latest/) provided on their main page. 
- Here's one [example](https://stackoverflow.com/questions/20381160/importing-a-flask-security-instance-into-my-views-module-breaks-my-webapp)

daenney:
> My point was flask-dance doesn't work right off the bat or when using the example provided on their main page.

I don't believe that's true. The example on the home page, the GitHub one, doesn't need Blinkers b/c it doesn't register any signals. It's only when you use one of https://flask-dance.readthedocs.io/en/latest/signals.html?highlight=signal which afaik the example doesn't, and neither does the GitHub blueprint internally, that you need singlas.

The Stack Overflow example has nothing to do with Flask-Dance, that's about Flask-Security.

alevis:
My apologies, was referring to the google [example.](https://flask-dance.readthedocs.io/en/latest/providers.html#module-flask_dance.contrib.google).

Which produced the same error in the stackoverflow example I referenced. 

But no worries. It’s  really just user preference.

singingwolfboy:
@alevis: The canonical Google example is here: [singingwolfboy/flask-dance-google](https://github.com/singingwolfboy/flask-dance-google). This example, and several others, are linked to from the [quickstart documentation](https://flask-dance.readthedocs.io/en/latest/quickstart.html#flask-sessions-easiest). Does that produce any errors? Because if so, you should file an issue in that repository.

PR #258, opened 2019-05-07
Provide Pytest fixture for recording HTTP with Betamax

singingwolfboy:
Writing automated tests for Flask-Dance is still hard. Hopefully this will make it a bit easier.


singingwolfboy:
I have no idea why this seems to make the test coverage drop so much. I suspect that it's a bug with `pytest-cov` or `coverage.py` somewhere, but I'm not yet sure...

singingwolfboy:
Yep, the problem was `pytest-cov`. When we were using that, it reported that this PR was dropping the project's coverage by over 15%, but when I switched to `coverage run`, it's now reporting that this PR is dropping project coverage by 0.15%, which is much more believable.

@daenney, what do you think of this change? Is the documentation understandable?

setup.py
```diff
@@ -63,10 +63,12 @@ def get_requirements(path):
     tests_require=get_requirements("dev-requirements.txt"),
     extras_require={"sqla": ["sqlalchemy", "sqlalchemy-utils"], "signals": ["blinker"]},
     cmdclass={"test": PyTest},
+    entry_points={"pytest11": ["pytest_flask_dance = flask_dance.fixtures.pytest"]},
```
singingwolfboy: This magical line allows others to use this fixture in their tests. [See the pytest docs on writing plugins.](https://docs.pytest.org/en/latest/writing_plugins.html#making-your-plugin-installable-by-others)

daenney: I was very confused about that line until I found the docs, the naming is fairly unobvious.

daenney:
I think my only concern would be that we don't explain that the first time betamax will record this test for you, and subsequent times it'll be loading the cassette. If I recall we do already mention that somewhere in the docs.

singingwolfboy:
I think we can assume the reader is able to follow links and read more about Betamax in the Betamax documentation. I'm also not sure how to make this any clearer, so I'm just going to merge this.

PR #261, opened 2019-05-08
google: Remove inline code example

daenney:
None of the other providers have one, and the official quickstarts provide a better point of reference.


singingwolfboy:
@atwalsh: you added this example, would you like to talk about it on this pull request? Maybe we can find a better place in the documentation for this code sample. It see the value of including this code in the documentation, but it does break up the flow of the page on https://flask-dance.readthedocs.io/en/latest/providers.html. Maybe we should break this up, into one page per provider?

daenney:
So there's 2 things here:
* It breaks up the flow
* It uses a signal which means this example breaks by default b/c blinkers is an optional dependency

atwalsh:
@singingwolfboy I have no problem with removing the code example. After looking at the providers page again I agree that the code block seems out of place. I think the warning provides enough info, and I could always reference the commit in any future issues. I vote that until another provider requires additional code examples, we keep this as one page and remove the `hosted_domain` code example.

Issue #263, opened 2019-05-22
Okta provider not working


RichardCullen:
Trying to use the Okta provider (which doesn't have an example to follow) 

```
import os
from flask import Flask, redirect, url_for
from flask_dance.contrib.okta import make_okta_blueprint, okta
from flask_dotenv import DotEnv

app = Flask(__name__)
env = DotEnv(app)

app.secret_key = os.environ.get("FLASK_SECRET_KEY", "supersekrit")
okta_bp = make_okta_blueprint(
    client_id=app.config["OKTA_OAUTH_CLIENT_ID"],
    client_secret=app.config["OKTA_OAUTH_CLIENT_SECRET"],)
app.register_blueprint(okta_bp, url_prefix="/login")


@app.route("/")
def index():
    if not okta.authorized:
        return redirect(url_for("okta.login"))
    resp = okta.get("/user")
    assert resp.ok
    return "You are @{login} on Okta".format(login=resp.json()["login"])


if __name__ == "__main__":
    app.run(debug=True, use_reloader=True)
```

Gives me the following error:
```
builtins.AttributeError
AttributeError: 'NoneType' object has no attribute 'lower'

Traceback (most recent call last)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 2328, in __call__
return self.wsgi_app(environ, start_response)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 2314, in wsgi_app
response = self.handle_exception(e)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 1760, in handle_exception
reraise(exc_type, exc_value, tb)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\_compat.py", line 36, in reraise
raise value
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 2311, in wsgi_app
response = self.full_dispatch_request()
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 1834, in full_dispatch_request
rv = self.handle_user_exception(e)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\app.py", line 1737, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "C:\work\python\okta-flask-example\env\lib\site-packages\flask\_compat.py", line 36, in reraise
raise value
[...]

PR #267, opened 2019-06-04
Fixed issues with Okta blueprint

tnolan8:
Finally got around to doing this. Apologies for the delay! Couple of things:

* As per discussion on #263 I've made `base_url` a required field for the blueprint.
* I've also made `authorization_url` and `token_url` required. The reason being that part of that URL is also configurable. In general they look like: <OKTA_BASE_URL>/oauth2/default/v1/{authorize|token} but the 'default' bit here is configuration and can be renamed to whatever you want. I could try come up with a way of trying to have a default value but that's also dependant on what's used for the BASE_URL.

Let me know if there's anything that needs to be changed! Again sorry about the delay, it's been pretty hectic. Also I created a sample app at: [flask-dance-okta](https://github.com/tnolan8/flask-dance-okta), not totally sure where to place that in the doc!

### DIFF
--- a/flask_dance/contrib/okta.py
+++ b/flask_dance/contrib/okta.py
@@ -14,15 +14,16 @@ __maintainer__ = "Tom Nolan <tomnolan95@gmail.com>"
 
 
 def make_okta_blueprint(
+    base_url,
+    authorization_url=None,
+    token_url=None,
     client_id=None,
     client_secret=None,
-    base_url=None,
     scope=None,
     redirect_url=None,
-    token_url=None,
     redirect_to=None,
     login_url=None,
-    authorization_url=None,
+    authorized_url=None,
     session_class=None,
     storage=None,
 ):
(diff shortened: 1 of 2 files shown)


singingwolfboy:
Rather than making `authorization_url` and `token_url` required, it would be better to set the defaults based on the `base_url`, [the way the GitLab configuration does](https://github.com/singingwolfboy/flask-dance/blob/master/flask_dance/contrib/gitlab.py#L66-L70). Also, remember that the automated tests need to pass.

tnolan8:
My bad forgot to update the test! Sure, I don't mind setting defaults 🙂

singingwolfboy:
@tnolan8: there's only so much customization we should allow in a pre-set configuration. The whole point of a pre-set configuration is that some of the configuration values are _set_ -- if you don't want that, then you should just use the `OAuth2ConsumerBlueprint` class directly. Maybe the correct way to support Okta is by providing some good documentation, rather than code.

How would you feel about _removing_ the Okta pre-set configuration, and adding some documentation and/or examples of how you could use Okta with Flask-Dance?

tnolan8:
Sure if you think that's best I can remove `okta.py` and it's associated tests and what not. I can update my sample app to use the `OAuth2ConsumerBlueprint`.

singingwolfboy:
Sounds good to me. You can add a link to your project to the `docs/quickstart.rst` file -- or if you prefer, you can write a dedicated page of the documentation for how to make a custom provider, and use Okta as an example. You may also want to check out the "Custom" section of the `docs/providers.rst` file -- it's at the very bottom. I'd be happy to find a time to videochat with you sometime, and discuss the best way to document this. It's a tricky concept, but an important one to have in the docs!

I'm going to close this pull request, and you can open a new one for removing the Okta pre-set configuration and modifying the docs.

PR #271, opened 2019-06-06
Adding Response object explanation to multi-user setup.

### LINKED ISSUE #270: Redirecting after OAuth dance AND using multi-user setup
Based on the documentation, when using [multi-user setup ](https://flask-dance.readthedocs.io/en/latest/multi-user.html) we're supposed to return False to disable Flask default behavior for saving the token. With the new update regarding #253 ,  now if we want to redirect, we need to do this after the `@oauth_authorized` signal.
```python
@oauth_authorized.connect
def logged_in(blueprint, token):
    blueprint.token = token
    next_url = flask.session["next_url"]
    return flask.redirect(next_url)
```

How can we return a flask.redirect and return False at the same time in the same method? 
It does not seem possible to me. Is there a new behaviour I need to know in order to disable the default token saving?

jmprovencher:
This adds clarification in the multi-user setup part of the documentation on how to override Flask-Dance default behaviour. In my application, I had to redirect the user to a dynamic url after creating his account, however I didn't understand how I could redirect the user and return False from the signal at the same time. I think this clarifies it. 

resolves #270

### DIFF
--- a/docs/multi-user.rst
+++ b/docs/multi-user.rst
@@ -114,8 +114,10 @@ behavior. To do so, you'll need to hook into the
 Flask-Dance's default behavior comes from storing the OAuth token for you
 automatically. To override the default behavior, write a function that
 subscribes to this signal, handles it the way *you* want,
-and returns ``False``. Returning ``False`` from this signal handler indicates
-to Flask-Dance that it should not try to store the OAuth token for you.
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Thanks! :tada:

PR #285, opened 2019-08-27
Update documentation around google env vars.

dc740:
I documented the changes needed to use custom environment variable names with the google blueprint. Please feel free to rephrase it, corrections are well received. I'm not a native english speaker, and it may need changes.

### DIFF
--- a/flask_dance/contrib/google.py
+++ b/flask_dance/contrib/google.py
@@ -30,10 +30,16 @@ def make_google_blueprint(
 ):
     """
     Make a blueprint for authenticating with Google using OAuth 2. This requires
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Thanks for doing this! However, this information is not specific to the Google blueprint factory, so I don't think this is the right place to put it. Maybe we can put this information with the `OAuth2ConsumerBlueprint` documentation, and link to it from the Google blueprint docs (and all the other providers as well). How does that sound?

singingwolfboy:
No response from the author in more than a year, so I'm closing this pull request.

PR #309, opened 2020-06-22
Update for Discord moving to discord.com

fenhl:
Since [discordapp.com](https://discordapp.com/) is now a redirect to [discord.com](https://discord.com/) which additionally shows a notice about the address change every time you use it, this updates the provider so the new domain is used and the notice should no longer appear.


daenney:
@singingwolfboy Our tests are broken I think. The change @fenhl made is unrelated to the test failures from what I can see.

singingwolfboy:
You're absolutely right. I spent some time today fixing up our tests, and moving over from Travis CI to GitHub Actions. @fenhl, could you please rebase your branch, so that it picks up those changes? The tests should pass after that.

fenhl:
Done.

singingwolfboy:
Thank you! 😄

PR #330, opened 2020-11-17
Add Digital Ocean as a provider

mikeabrahamsen:
- add digitalocean.py to configure blueprint for provider
- add tests for Digital Ocean provider
- update CHANGELOG
- update docs/providers.rst
- add DigitalOcean to wordlist

Question:
The default scope with Digital Ocean is `read`. I left the default as `None` for library consistency, would it make sense in this case to change the default, or is this something we want the user to explicitly set?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Changelog
 
 `unreleased`_
 -------------
-nothing yet
+Added Digital Ocean pre-set configuration
 
 `3.1.0`_ (2020-10-29)
 ---------------------
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -36,7 +36,7 @@ Azure
     has authenticated with Azure AD at some point in the past).
 
 Digital Ocean
------
+-------------
 .. module:: flask_dance.contrib.digitalocean
 
 .. autofunction:: make_digitalocean_blueprint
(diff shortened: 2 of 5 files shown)


daenney:
We have at least 3 providers in `contrib` with a default for `scope` that's not `None`, including the recently added one for Strava. If having a scope of `None` doesn't really make sense for the DO provider, I'm happy to see that changed to `read`.

@singingwolfboy Do you have a preference here? It does feel like we're being a bit inconsistent by some providers having a default scope we set, whereas most have no scope at all. I personally prefer the idea of explicitly having to ask for scopes when using the `make_*_blueprint` functions, but there's value in a sensible default too.

mikeabrahamsen:
I prefer explicitly setting the scope as well. I created an example repo of how to use this provider here: https://github.com/mikeabrahamsen/flask-dance-digitalocean-oauth-example

singingwolfboy:
> It does feel like we're being a bit inconsistent by some providers having a default scope we set, whereas most have no scope at all.

Good point. However, that feels different from this specific provider. Let's open a separate issue to discuss the inconsistency across providers.

@mikeabrahamsen This looks great! Thank you for doing this :-)

singingwolfboy:
This is now published in version 3.2.0.

PR #332, opened 2020-11-27
Add support for Dropbox short-lived offline refresh tokens.

thunderkeys:
This patch adds support for offline refresh tokens for Dropbox, which will be required beginning September 2021.

https://dropbox.tech/developers/migrating-app-permissions-and-access-tokens

### DIFF
--- a/flask_dance/consumer/requests.py
+++ b/flask_dance/consumer/requests.py
@@ -145,6 +145,13 @@ class OAuth2Session(BaseOAuth2Session):
         return self.token and self.token.get("access_token")
 
     @property
+    def refresh_token(self):
+        """
+        Returns the ``refresh_token`` from the OAuth token.
+        """
+        return self.token and self.token.get("refresh_token")
+
+    @property
     def authorized(self):
         """This is the property used when you have a statement in your code
         that reads "if <provider>.authorized:", e.g. "if twitter.authorized:".
--- a/flask_dance/contrib/dropbox.py
+++ b/flask_dance/contrib/dropbox.py
@@ -17,6 +17,7 @@ def make_dropbox_blueprint(
     app_key=None,
     app_secret=None,
     scope=None,
+    offline=False,
     force_reapprove=False,
     disable_signup=False,
     require_role=None,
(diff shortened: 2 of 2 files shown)


flask_dance/consumer/requests.py
```diff
@@ -144,6 +144,13 @@ def access_token(self):
         """
         return self.token and self.token.get("access_token")
 
+    @property
+    def refresh_token(self):
```
singingwolfboy, line 148: Modifying the `OAuth2Session` class seems like a larger change than supporting this change from Dropbox. Can you move it to a separate pull request? It's also going to need an automated test and updated documentation to support it.

singingwolfboy (changes requested):
Hey @thunderkeys, thanks for the pull request! Sorry it took me so long to review it. :-(

We're going to need an automated test for this change before it can get merged. If you're not sure how to write one, I'm happy to help you with that if you want -- maybe we could find a time to videochat and paircode or something. :-) The change to `OAuth2Session` should also be its own pull request, rather than rolled into this one. Once you do those things, I think this can be merged!

Also, are you interested in being the maintainer for the Dropbox configuration? I don't actively use it these days, so it makes sense to have the maintainer be someone who _is_ using it.

rooterkyberian:
@thunderkeys thanks for being vigilant about this - I would have missed if it wasn't for you.

Do you think you will find time to update your PR or can I recreate it addressing the reviewer comment so we can move forward and (hopefully) get it released?

In the mean time I'm using following workaround:
```
auth_blueprint = make_dropbox_blueprint(...)
auth_blueprint.authorization_url_params["token_access_type"] = "offline"
```

thunderkeys:
Hi - I don't mind if you recreate it with the suggestions, it may be a bit before I have time to.  Good idea on the workaround.

flask_dance/consumer/requests.py
rooterkyberian: this part of the PR is only thing not yet merged

rest was merged in some form in #339

singingwolfboy:
Implemented by #339.

PR #334, opened 2020-12-26
Added Discord prompt param

nerrixDE:
As documented [here](https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-authorization-url-example) this would be good to get implemented.

### DIFF
--- a/tests/contrib/test_discord.py
+++ b/tests/contrib/test_discord.py
@@ -39,6 +39,24 @@ def test_blueprint_factory():
     assert discord_bp.token_url == "https://discord.com/api/oauth2/token"
 
 
+def test_blueprint_factory_with_prompt():
+    discord_bp = make_discord_blueprint(
+        client_id="foo",
+        client_secret="bar",
+        scope=["identify", "email"],
+        redirect_to="index",
+        prompt=None,
+    )
+    assert isinstance(discord_bp, OAuth2ConsumerBlueprint)
+    assert discord_bp.session.scope == ["identify", "email"]
+    assert discord_bp.session.base_url == "https://discord.com/"
+    assert discord_bp.session.client_id == "foo"
+    assert discord_bp.client_secret == "bar"
+    assert discord_bp.authorization_url == "https://discord.com/api/oauth2/authorize"
+    assert discord_bp.token_url == "https://discord.com/api/oauth2/token"
+    assert discord_bp.authorization_url_params["prompt"] == None
+
+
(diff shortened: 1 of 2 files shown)


daenney:
I don't think this should default to `consent`. If the user hasn't authorized previously or the scopes have changed, even with a prompt of `None` Discord will still prompt. A default of `consent` will always cause Discord to ask even without a scope change, which is not the experience most people want or expect out of the box.

I'd probably change the implementation to do what we do in the Google provider, with a `reprompt_consent` that defaults to `False` but can be set to `True` and then an `if reprompt_consent` add it to the `authorization_url`.

nerrixDE:
@daenney I agree, however Discord's default is `consent` and right now it is set to `consent` implicitly. Thus setting it to `None` on default would change existing behaviour (There is only `None` and `consent`)

daenney:
Ah I see. It wasn't clear to me from the docs that if it's omitted they default it to consent instead of none.

I'd still like to see this changed to take a boolean of  `reprompt_consent` instead, to keep it inline with the 2 other providers where we do this (Google and Zoho), but in that case default it to `True` instead to keep the current behaviour intact.

nerrixDE:
@daenney Good point however I would personally prefer to keep the field as it is defined by the API (Azure provider uses `prompt` as well). Also putting a boolean wrapper around it may cause issues in case Discord decides to add additional prompt-modes in future.

daenney:
Fair point, I had missed the Azure one.

daenney:
Can you add an additional test though? Right now only the behaviour of explicitly setting to to `None` is tested, which is no longer the default of the blueprint.

tests/contrib/test_discord.py
```diff
@@ -39,6 +39,24 @@ def test_blueprint_factory():
     assert discord_bp.token_url == "https://discord.com/api/oauth2/token"
```
daenney, line 40: You should add an `assert` to `test_blueprint_factory()` too, to ensure the default prompt value of consent is now there.

tests/contrib/test_discord.py
```diff
@@ -66,6 +84,7 @@ def test_context_local(make_app):
         "bar2",
         redirect_to="url2",
         storage=MemoryStorage({"access_token": "app2"}),
+        prompt=None,
```
daenney, line 87: I don't think this one is necessary anymore?

daenney (approved):

PR #335, opened 2020-12-26
test: Fix codecov reporting

daenney:
By default, `coverage run` does not generate a `coverage.xml`, you have to call `coverage xml` for that. Since we're not doing that, `coverage.xml` doesn't exist so we don't end up uploading code coverage.

It turns out however that doing this is unnecessary. The codecov action will do this by itself if it has to, as long as it finds a `.coverage` which coverage.py will write to by default.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,6 @@ jobs:
 
       - name: Upload coverage to Codecov
         if: ${{ always() }}
-        uses: codecov/codecov-action@v1.0.7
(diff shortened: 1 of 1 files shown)


daenney:
@singingwolfboy I went ahead and merged it, hope that's OK. I noticed our coverage was broken.

daenney:
Sigh. Annoyingly we're now below the coverage target for main, so GitHub shows a nice ❌ next to the commit even though CI passed just fine.

PR #340, opened 2021-02-03
Drop EoLed Python versions (including python 2)

rooterkyberian:
All python versions older than 3.6 are long dead.
https://endoflife.date/python

Motivation for removing them is quite simple - its just easier to support less python versions and python 3.x line has introduced number of nice features that we can start leveraging (fstrings, type hints among others). You can already see some improvements done by `pyupgrade` in this PR.

Please note just removing them here, thanks to `pip` support for [`python_requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) will not break old installations assuming they have a  setuptools>=24.2.0 and pip>=9.0.0 .

1) document the python versions supported in setup.py
2) update tox, github actions config
3) run `pyupgrade --py36-plus`
4) remove `six` package

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 requests>=2.0
 oauthlib
 requests-oauthlib>=1.0.0
-Flask>=0.7
+Flask>=0.9
 urlobject
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@ oauthlib
 requests-oauthlib>=1.0.0
 Flask>=0.7
 urlobject
-six
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["2.x", "3.x"]
+        python-version: ["3.6", "3.x"]
         database-uri:
           - "postgresql://testuser:testpw@localhost/testdb"
           - "sqlite://"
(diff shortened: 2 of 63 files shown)


daenney:
I'm generally fine with this and I agree it makes some stuff easier. I'd like to get the Dropbox one landed and tagged with Python 2 support still in there though. After that we can proceed with this PR and tag a new major version of the library.

@singingwolfboy Do you have any other big things you'd like to do? I don't feel we can drop Python 2 support in a minor release, so now would be an opportunity to do any other changes that might not be backwards compatible.

singingwolfboy:
I just released v3.3.0, and I think we can make this change for Flask-Dance v4.0.0. This PR is already pretty comprehensive; the only thing I can think of off the top of my head is simplifying all the places where we do this:

```python
try:
    from flask import _app_ctx_stack as stack
except ImportError:
    from flask import _request_ctx_stack as stack
```

We should be able to just do `from flask import _app_ctx_stack` -- the `except ImportError` block is only relevant for very, very old versions of Flask that might not even run on Python 3.

rooterkyberian:
rebased to current `main`, removed old flask import

requirements.txt
```diff
@@ -3,4 +3,3 @@ oauthlib
 requests-oauthlib>=1.0.0
 Flask>=0.7
```
daenney, line 4: Should we maybe up the Flask requirement to 1.0? It was released in April 2018. It's made it into most distributions by now.

rooterkyberian: the change done in this ticket requires only Flask>=0.9

rooterkyberian:
anything else?

singingwolfboy:
Thanks so much! Just merged it 🎉

PR #342, opened 2021-02-12
Implemented atlassian blueprint for oauth2

Przemoo16:
According to https://www.atlassian.com/migration/faqs#general, Atlassian stopped selling and is going to stop maintaining server products and focus on cloud products which use Oauth2 authentication to allow external applications and services to access Atlassian product APIs.

Please suggest naming changes since there already is jira (oauth1) plugin.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 `unreleased`_
 -------------
 * Add `offline` option to `make_dropbox_blueprint`
+* Added Atlassian pre-set configuration
 
 `3.2.0`_ (2020-11-24)
 ---------------------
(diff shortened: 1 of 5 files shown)


daenney:
Thanks you for the contribution! Unfortunately we can't accept a PR without some basic tests to ensure things don't regress. Please take a look at the different tests in https://github.com/singingwolfboy/flask-dance/tree/main/tests/contrib for inspiration.

daenney:
On the naming front, I would suggest naming it 'atlassian' since you're authenticating for any/all Atlassian Cloud products.

daenney (approved):
This looks good! I'll leave it open for a day or two so other folks can review and otherwise I'll merge it.

singingwolfboy:
This is still missing a few points on the checklist here: https://github.com/singingwolfboy/flask-dance/blob/main/flask_dance/contrib/README.rst

In particular, we need a `__maintainer__` variable, the docs need to be updated, and there must be an entry in the changelog.

PR #346, opened 2021-02-26
Added domain option to Salesforce blueprint

Przemoo16:
Right now Salesforce will be working only with the production instance. That change adds the possibility to specify custom domains in the Salesforce blueprint. 

Also changed base URL in the blueprint. It will make easier to use the session in further requests.

### DIFF
--- a/flask_dance/contrib/salesforce.py
+++ b/flask_dance/contrib/salesforce.py
@@ -18,6 +18,8 @@ def make_salesforce_blueprint(
     client_secret=None,
     scope=None,
     reprompt_consent=False,
+    hostname=None,
+    is_sandbox=False,
     redirect_url=None,
     redirect_to=None,
     login_url=None,
@@ -39,6 +41,11 @@ def make_salesforce_blueprint(
         reprompt_consent (bool): If True, force Salesforce to re-prompt the user
             for their consent, even if the user has already given their
             consent. Defaults to False.
+        hostname (str, optional): The hostname of your Salesforce instance.
+            By default, Salesforce uses ``login.salesforce.com`` for production
+            instances and ``test.salesforce.com`` for sandboxes.
+        is_sandbox (bool): If hostname is not defined specify whether Salesforce
+            instance is a sandbox. Defaults to False.
         redirect_url (str): the URL to redirect to after the authentication
             dance is complete.
         redirect_to (str): if ``redirect_url`` is not defined, the name of the
@@ -62,15 +69,20 @@ def make_salesforce_blueprint(
     if reprompt_consent:
         authorization_url_params["prompt"] = "consent"
 
+    if not hostname:
+        hostname = "test.salesforce.com" if is_sandbox else "login.salesforce.com"
+
     salesforce_bp = OAuth2ConsumerBlueprint(
         "salesforce",
         __name__,
         client_id=client_id,
         client_secret=client_secret,
         scope=scope,
-        base_url="https://login.salesforce.com/services/oauth2/",
-        authorization_url="https://login.salesforce.com/services/oauth2/authorize",
-        token_url="https://login.salesforce.com/services/oauth2/token",
+        base_url="https://{hostname}/".format(hostname=hostname),
+        authorization_url=(
+            "https://{hostname}/services/oauth2/authorize".format(hostname=hostname)
+        ),
+        token_url="https://{hostname}/services/oauth2/token".format(hostname=hostname),
         redirect_url=redirect_url,
         redirect_to=redirect_to,
         login_url=login_url,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,9 @@ Changelog
 
 `unreleased`_
 -------------
-nothing yet
+* Added `hostname` option to the `make_salesforce_blueprint`
+* Added `is_sandbox` option to the `make_salesforce_blueprint`
+* Changed base url for `make_salesforce_blueprint`
 
 `3.3.0`_ (2021-02-25)
 ---------------------
--- a/tests/contrib/test_salesforce.py
+++ b/tests/contrib/test_salesforce.py
@@ -24,7 +24,7 @@ def make_app():
     return _make_app
 
 
-def test_blueprint_factory():
+def test_blueprint_factory_default():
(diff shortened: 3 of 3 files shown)


daenney (approved):

flask_dance/contrib/salesforce.py
```diff
@@ -62,15 +64,25 @@ def make_salesforce_blueprint(
...
     salesforce_bp = OAuth2ConsumerBlueprint(
         "salesforce",
         __name__,
         client_id=client_id,
         client_secret=client_secret,
         scope=scope,
-        base_url="https://login.salesforce.com/services/oauth2/",
-        authorization_url="https://login.salesforce.com/services/oauth2/authorize",
-        token_url="https://login.salesforce.com/services/oauth2/token",
+        base_url="https://{subdomain}.salesforce.com/services/oauth2/".format(
```
singingwolfboy, line 75: Maybe this is a dumb question, but does Salesforce support any _other_ subdomains as well? Or only `test` and `login`?

Przemoo16: Salesforce has two types of environments: production and sandbox (testing). The authorization URLs (refreshing token URL as well) are different depending on the environment so production environments have `login` subdomain and sandbox environments have `test` one.

Przemoo16: I have checked it more precisely and your question was accurate. It is possible to create custom domains on the Salesforce so I am going to change that PR to handle any subdomain.

flask_dance/contrib/salesforce.py
```diff
@@ -39,6 +40,9 @@ def make_salesforce_blueprint(
         reprompt_consent (bool): If True, force Salesforce to re-prompt the user
             for their consent, even if the user has already given their
             consent. Defaults to False.
+        domain (str, optional): The domain of your Salesforce instance. By default,
```
singingwolfboy, line 43: Does the Salesforce documentation refer to this as a "domain" or a "subdomain"? Technically, this is a subdomain of the `salesforce.com` domain, but we should probably follow whatever terminology is in the Salesforce documentation. I have never used the Salesforce API, so I leave this up to you.

flask_dance/contrib/salesforce.py
```diff
@@ -68,9 +72,15 @@ def make_salesforce_blueprint(
         client_id=client_id,
         client_secret=client_secret,
         scope=scope,
-        base_url="https://login.salesforce.com/services/oauth2/",
-        authorization_url="https://login.salesforce.com/services/oauth2/authorize",
-        token_url="https://login.salesforce.com/services/oauth2/token",
+        base_url="https://{domain}.salesforce.com/".format(domain=domain),
```
singingwolfboy, line 75: Thanks for mentioning this in the changelog. This _seems_ like a more reasonable base URL to me, but can you verify that API calls to Salesforce all generally start with this URL? Just want to make sure that this is the correct value, since the original value was apparently wrong.

singingwolfboy (changes requested):

flask_dance/contrib/salesforce.py
Przemoo16: You can make an authorization process in two ways:
1) Using your instance URL e.g `example.my.salesforce.com/services/oauth2/authorize`
2) Without any information about the instance url: `login.salesforce.com/services/oauth2/authorize` (production), `test.salesforce.com/services/oauth2/authorize` (sandbox)

When it comes to API calls you need to know your instance URL and make calls using that url e.g. `example.my.salesforce.com/services/data/...`

At that moment when the user specifies his domain as `example.my` he can make API calls using url from the previous example. Otherwise, the default URL will be used (login.salesforce.com), and he cannot do that.

I don't want to make the user to specifies his instance URL obligatory because he can be authorized using defaults endpoints and later fetch his instance url.

If you have any other ideas I am open to discuss that :)

flask_dance/contrib/salesforce.py
Przemoo16: Yes this is a subdomain but leaded me to that decision was that Salesforce uses `My Domain` naming for possibility to customize instance's URL (https://help.salesforce.com/articleView?id=sf.domain_name_overview.htm&type=5). It's not clear so if you think the subdomain will be more clear to the user (domain can suggest inserting full domain name) I can change it. But keep in mind that user needs to put sometimes more subdomains like (example.my) so subdomain naming can also be misleading.

flask_dance/contrib/salesforce.py
singingwolfboy: `base_url` is not used for the authorization process, but for making API calls _after_ the OAuth token is obtained. For example, with the GitHub blueprint, you can run `github.get("/user")`, and it will access the URL `https://api.github.com/user`, because the `base_url` is `https://api.github.com/`.

Can you point me to some documentation for how to make API calls to Salesforce, _after_ you've already gotten an OAuth token? That will help to determine the correct value for `base_url`.

Przemoo16: Yea I know but to make API calls you need to know your instance URL, but for authorization you don't. So I can leave the base_url as None if the user hasn't provided domain. Sample calls: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/quickstart_code.htm

Przemoo16:
Another approach which gives the user better understanding parameters and more flexibility: https://github.com/singingwolfboy/flask-dance/pull/346/commits/db8ad30fefcb61bb6cc7370d30a98d9ac3b807cf

flask_dance/contrib/salesforce.py
Przemoo16: It solves the issue of domain vs subdomain: https://github.com/singingwolfboy/flask-dance/pull/346/commits/db8ad30fefcb61bb6cc7370d30a98d9ac3b807cf

singingwolfboy (approved):
I like this `hostname` argument a lot more than `domain`. Good idea!

> But still the question about base_url. Should we leave it empty if user doesn't define hostname?

At that point in time, `hostname` will always be defined, right? Because of the logic around the `is_sandbox` parameter.

Any other changes you want to make? Or is this ready to merge?

Przemoo16:
Yes `hostname` will always be defined. I squashed the commits and now it is ready to merge.

Issue #123, opened 2018-04-30
Flask dance with twitch API


chanux:
I have been trying to use flask-dance for Twitch API via OAuth2ConsumerBlueprint. The requests seem to fail since Twitch API expects client ID in request headers.

```DEBUG:requests_oauthlib.oauth2_session:Requesting url https://api.twitch.tv/user using method GET.
DEBUG:requests_oauthlib.oauth2_session:Supplying headers {u'Authorization': u'Bearer XXXXXX'} and data None
DEBUG:requests_oauthlib.oauth2_session:Passing through key word arguments {'allow_redirects': True}.
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.twitch.tv
DEBUG:urllib3.connectionpool:https://api.twitch.tv:443 "GET /user HTTP/1.1" 302 154
DEBUG:urllib3.connectionpool:https://api.twitch.tv:443 "GET /kraken/base HTTP/1.1" 400 96
{u'status': 400, u'message': u'No client id specified', u'error': u'Bad Request'}
```

I have tried manually setting the headers and it seems to succeed  (even though it leaves new questions for me). I can not see how I can set custom headers. Is it possible?

singingwolfboy:
Flask-Dance is built on top of [Requests](http://python-requests.org/). If you want to set custom headers, [you can do so the same way that you normally do with Requests](http://docs.python-requests.org/en/master/user/quickstart/#custom-headers). You can also set default headers on the [Session object](http://docs.python-requests.org/en/master/user/advanced/#session-objects).

Does that answer your question? If you get it working, it would be great if you could contribute a Twitch configuration for Flask-Dance, so others can use it. :smile:

chanux:
I have tried to update the headers on session object. Still no luck. Maybe I'm doing something wrong :). I'd love to contribute a twitch blueprint if I managed get things to work.

Documenting the fruit of my misery so far to save precious time for the others who have to go through this 😆 


```python
app = Flask(__name__)
twitch_blueprint = OAuth2ConsumerBlueprint(
    "twitch", __name__,
    client_id='xxxx',
    client_secret='xxxx',
    base_url="https://api.twitch.tv/kraken",
    token_url="https://id.twitch.tv/oauth2/token",
    authorization_url="https://id.twitch.tv/oauth2/authorize",
    redirect_url="http://localhost:8080/authenticated",
    scope=('openid', 'user:read:email'),
)
[...]

PR #348, opened 2021-04-03
Add Twitch Blueprint

kerryhatcher:
initial import from custom code.

I'm really not a fan of how the `client-id` header is getting set in `session_created()` The id isn't available in the session object when the function runs. If anyone has a better way, that would be great. 

I'm not sure if all the docs & tests are there so I would really appreciate some help with that.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ sphinxcontrib-spelling
 Flask-Sphinx-Themes
 
 # code dependencies, needed for imports
-sqlalchemy>=0.9
+sqlalchemy>=0.9,<1.4.0
 sqlalchemy-utils
 pytest
 betamax
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,6 @@ sphinxcontrib-spelling
 Flask-Sphinx-Themes
 
 # code dependencies, needed for imports
-sqlalchemy>=0.9,<1.4.0
-sqlalchemy-utils
+sqlalchemy>=1.3.11
 pytest
 betamax
--- a/flask_dance/contrib/twitch.py
+++ b/flask_dance/contrib/twitch.py
@@ -6,7 +6,6 @@ import os
 from flask import _app_ctx_stack as stack
 
 
-
 __maintainer__ = "Kerry Hatcher <kerry.hatcher@protonmail.com>"
 
 
@@ -14,9 +13,11 @@ class TwitchBlueprint(OAuth2ConsumerBlueprint):
     """
     Add client-id to the API request headers
     """
+
     def session_created(self, session):
         session.headers.update({"client-id": os.getenv("TWITCH_CLIENTID")})
         return session
+
 
 def make_twitch_blueprint(
     client_id=None,
@@ -57,6 +58,7 @@ def make_twitch_blueprint(
     :rtype: :class:`~flask_dance.consumer.OAuth2ConsumerBlueprint`
     :returns: A :ref:`blueprint <flask:blueprints>` to attach to your Flask app.
     """
+
     twitch_bp = TwitchBlueprint(
         "twitch",
         __name__,
--- a/flask_dance/contrib/twitch.py
+++ b/flask_dance/contrib/twitch.py
@@ -93,22 +93,12 @@ def make_twitch_blueprint(
 
         if twitch_bp.client_id is None:
             raise MissingConfigException(
-                f"client_id is required, currently is {type(twitch_bp.client_id)}"
+                f"client_id is required, currently is {twitch_bp.client_id}. Type: {type(twitch_bp.client_id)}"
             )
 
-        if type(twitch_bp.client_id) != str:
+        if twitch_bp.client_secret is None:
             raise MissingConfigException(
-                f"client_id should be a str, is: {type(twitch_bp.client_id)}"
-            )
-
-        if twitch_bp.client_id is None:
-            raise MissingConfigException(
-                f"client_id is required, currently is {type(twitch_bp.client_id)}"
-            )
-
-        if type(twitch_bp.client_secret) != str:
-            raise MissingConfigException(
-                f"client_secret should be a str, is {type(twitch_bp.client_secret)}"
+                f"client_secret is required, currently is  is {twitch_bp.client_secret}. Type: {type(twitch_bp.client_secret)}"
             )
 
         twitch_bp.session.headers.update({"client-id": twitch_bp.client_id})
--- a/flask_dance/contrib/twitch.py
+++ b/flask_dance/contrib/twitch.py
@@ -8,11 +8,11 @@ from flask import _app_ctx_stack as stack
 
 __maintainer__ = "Kerry Hatcher <kerry.hatcher@protonmail.com>"
 
+
 class TwitchBlueprint(OAuth2ConsumerBlueprint):
     def session_created(self, session):
-        session.headers.update({'client-id': os.getenv('TWITCH_CLIENTID')}) 
+        session.headers.update({"client-id": os.getenv("TWITCH_CLIENTID")})
         return session
-
 
 
 def make_twitch_blueprint(
--- a/flask_dance/contrib/twitch.py
+++ b/flask_dance/contrib/twitch.py
@@ -1,3 +1,5 @@
+"""Blueprint for connecting to Twitch API."""
+
 from flask_dance.consumer import OAuth2ConsumerBlueprint
 from functools import partial
 from flask.globals import LocalProxy, _lookup_app_object
@@ -8,15 +10,8 @@ from flask import _app_ctx_stack as stack
 
 __maintainer__ = "Kerry Hatcher <kerry.hatcher@protonmail.com>"
 
-
-class TwitchBlueprint(OAuth2ConsumerBlueprint):
-    """
-    Add client-id to the API request headers
-    """
-
-    def session_created(self, session):
-        session.headers.update({"client-id": os.getenv("TWITCH_CLIENTID")})
-        return session
+class MissingConfigException(Exception):
+    """Exception raised when a required config is missing."""
 
 
 def make_twitch_blueprint(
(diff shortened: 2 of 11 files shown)


docs/requirements.txt
```diff
@@ -4,7 +4,7 @@ sphinxcontrib-spelling
 Flask-Sphinx-Themes
 
 # code dependencies, needed for imports
-sqlalchemy>=0.9
+sqlalchemy>=0.9,<1.4.0
```
daenney, line 7: This doesn't look like it should be part of this PR

tests/requirements.txt
```diff
@@ -3,7 +3,7 @@ responses
 freezegun
 coverage
 # testing sqlalchemy support
-sqlalchemy>=0.9
+sqlalchemy>=0.9,<1.4.0
```
daenney, line 6: Same as https://github.com/singingwolfboy/flask-dance/pull/348/files#r607716737 ?

flask_dance/contrib/twitch.py
```diff
@@ -0,0 +1,91 @@
...
+__maintainer__ = "Kerry Hatcher <kerry.hatcher@protonmail.com>"
+
+
+class TwitchBlueprint(OAuth2ConsumerBlueprint):
+    """
+    Add client-id to the API request headers
+    """
+
+    def session_created(self, session):
+        session.headers.update({"client-id": os.getenv("TWITCH_CLIENTID")})
```
singingwolfboy, line 18: Rather than making this pull from an environment variable, it should pull from `app.config`. You can use the [`current_app`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.current_app) proxy to make that easier.

I'm also not sure if it makes sense to do this in the `session_created()` callback, or if it makes sense to override `request()` and do it there. The former happens only once, and if the configured value changes, it won't pick it up; the latter will always pull it for every request, so it will pick up updates.

Regardless, this _definitely_ needs to be covered in the automated tests. Your tests need to set this value in the app config, make an API request, and assert that the header is properly set. You should also think about what should happen if this value is not provided. If Twitch won't allow you to use their API without this header, then I suggest throwing an exception if this value is not provided. You should have an automated test for that, too.

docs/requirements.txt
singingwolfboy: I fixed the problems with the automated tests. @kerryhatcher, if you rebase on top of the latest version of the `main` branch, it should work!

flask_dance/contrib/twitch.py
kerryhatcher: YES! That makes a ton of sense!! Thanks for the guidance, I jumped straight in the deep end on this. Give me a day or so and I'll get all the changes in. 

As far as overriding `request()` goes I don't see an obvious use case for the client-id to change in a running app but that would be the most flexible for sure. Do you happen to have an example of that in another provider?

kerryhatcher: I think I got it... resolving this for now.

flask_dance/contrib/twitch.py
```diff
@@ -0,0 +1,111 @@
...
+        token_url=token_url,
+        token_url_params=token_url_params,
+        redirect_url=redirect_url,
+        redirect_to=redirect_to,
+        login_url=login_url,
+        authorized_url=authorized_url,
+        session_class=session_class,
+        storage=storage,
+    )
+    if twitch_bp.client_id is None:
```
singingwolfboy, line 76: You don't need these `if` statements here. The `from_config` values will only be used if the values are set in the app config; otherwise, they will be ignored. See https://github.com/singingwolfboy/flask-dance/blob/main/flask_dance/consumer/base.py#L80

flask_dance/contrib/twitch.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if twitch_bp.client_id is None:
+            raise MissingConfigException(
+                f"client_id is required, currently is {twitch_bp.client_id}. Type: {type(twitch_bp.client_id)}"
+            )
+
+        if twitch_bp.client_secret is None:
+            raise MissingConfigException(
+                f"client_secret is required, currently is  is {twitch_bp.client_secret}. Type: {type(twitch_bp.client_secret)}"
+            )
```
singingwolfboy, line 102: Is there a reason that you're doing this here? None of the other configurations behave this way -- which doesn't mean that this behavior is wrong, necessarily! I just want to understand why you're doing it. If we _are_ going to define a `MissingConfigException` class, it should probably be defined more broadly than just this specific configuration.

flask_dance/contrib/twitch.py
```diff
@@ -0,0 +1,111 @@
...
+
+        twitch_bp.session.headers.update({"client-id": twitch_bp.client_id})
```
singingwolfboy, line 104: Ah, so Twitch wants the OAuth client ID to appear in every API call? Interesting... and not compliant with the spec, as far as I know, but that's not unusual. Out of curiousity, can you link me to some documentation for their API that requests this?

flask_dance/contrib/twitch.py
```diff
@@ -0,0 +1,111 @@
...
+    client_id=None,
+    client_secret=None,
+    scope=None,
+    redirect_url=None,
+    redirect_to=None,
+    login_url=None,
+    base_url="https://api.twitch.tv/helix/",
+    authorization_url="https://id.twitch.tv/oauth2/authorize",
+    token_url="https://id.twitch.tv/oauth2/token",
+    token_url_params={"include_client_id": True},
```
singingwolfboy, line 28: These seem like too many options. The whole point of using a pre-set configuration is so that people don't _have_ to make as many choices. If a developer wants to specify all these options, she should just import and use the `OAuth2ConsumerBlueprint` class directly, rather than importing and using this configuration.

You can safely assume that anyone calling this function wants to make a blueprint that is specific to Twitch, so please remove any options that could allow them to make it work for any other OAuth provider.

singingwolfboy (changes requested):

flask_dance/contrib/twitch.py
kerryhatcher: I just thought it was a tad more descriptive than `TypeError` Its just something I started doing many moons ago after banging my head on a unhanded exception in a dependent library. Probably better to get included at a lower level. I can swap it out with `TypeError` easy enough. Its not so much that the type is wrong but rather missing. I took a glance at the python [error types ](https://docs.python.org/3/library/exceptions.html)and didn't see anything that really fit well. I'll take it in whatever direction you would like.

flask_dance/contrib/twitch.py
kerryhatcher: You can see it in the very first example here: https://dev.twitch.tv/docs/api/

I've not done an exhaustive search on every call but every one that I've looked at or have used requires it: 
https://dev.twitch.tv/docs/api/reference#start-commercial
https://dev.twitch.tv/docs/api/reference#search-channels
https://dev.twitch.tv/docs/api/reference#check-user-subscription

flask_dance/contrib/twitch.py
kerryhatcher: Roger, I think I understand.

flask_dance/contrib/twitch.py
kerryhatcher: Perfect!

flask_dance/contrib/twitch.py
singingwolfboy: Seems like a reasonable idea, but I think it's not related to adding the Twitch blueprint. As a result, it belongs in a separate pull request, so that we can discuss this idea separately -- and so that it can be applied to the whole codebase in a consistent way.

Please remove this `MissingConfigException` class from this pull request, and don't do this error checking here, to make it consistent with the other configurations. However, opening a new pull request that is focused only on adding a `MissingConfigException` class and applying it everywhere would be _very_ welcome!

kerryhatcher:
@singingwolfboy I think this about does it all other than the custom exception discussion.

kerryhatcher:
Also this will probably help folks like me who came here via https://github.com/singingwolfboy/flask-dance/issues/123 and couldn't figure out how folks had gotten the Twitch API to work in custom code.

flask_dance/contrib/twitch.py
singingwolfboy: These options are still here. Did you forget to remove them?

flask_dance/contrib/twitch.py
singingwolfboy: Here's an announcement about this: https://blog.twitch.tv/en/2016/05/05/client-id-required-for-kraken-api-calls-afbb8e95f843/

singingwolfboy (changes requested):

flask_dance/contrib/twitch.py
kerryhatcher: Will do! However I'm a little confused about your previous comment about throwing an exception. Are you suggesting that we just skip over that for now and come back to it in the separate PR? How should I handle the test for it if I'm not going to throw an error?  Sorry to be so needy in all this and thank you for being fantastically helpful!

>If Twitch won't allow you to use their API without this header, then I suggest throwing an exception if this value is not provided. You should have an automated test for that, too.

flask_dance/contrib/twitch.py
kerryhatcher: I can see why they would do that on the back end. Using the client-id as a primary key would make data-store lookup way faster. I'm sure they have a _few users_ out there. I'll mark this as resolved for now. Thanks!

flask_dance/contrib/twitch.py
singingwolfboy: Ah, that's a fair question. I asked you to do that when you were loading an environment variable in the `session_created()` function, and of course, environment variables could be defined or not defined. That's why I wanted you to raise an exception at that point: in case the environment variable was not defined. Also, this was an unusual case that was specific to the Discord configuration.

Now the situation has changed, and we're talking about something that is more broadly applicable to the entire codebase. In addition, we're no longer talking about a separate environment variable; we're talking about `twitch_bp.client_id` and `twitch_bp.client_secret`, which _must_ be defined; if they're not, the system will break in other, more interesting ways. (That's why I think the idea of raising a dedicated `MissingConfigException` is a good idea, but one that should be applied more globally.)

As a result, we don't need a test for what happens when these values are not defined -- at least, not until that other `MissingConfigException` pull request is created. We _do_ still need to test that the `Client-Id` header is applied to API requests to Twitch, but you've already done that, so I think we're good. 👍

singingwolfboy:
Hi @kerryhatcher, I created a new pull request based off of this one (#356). I added a few changes and merged it, so that we could get Flask-Dance 4.0 shipped. Thanks for the pull request!

kerryhatcher:
@singingwolfboy Awesome! I had to go do military stuff and just got back. Glad I could help! I'm going to work on the other things we discussed sometime this coming week assuming I don't get called up again.

PR #353, opened 2021-04-08
Send specific kwargs to Oauth2CustomerBlueprint

willpwa:
Due to the additional setup in this function, it is preferable to use `make_gitlab_blueprint` instead of creating a `Oauth2CustomerBlueprint` directly. However, in the event of a self-hosted gitlab instance using self-signed certificates, these certificates cannot always be verified and this makes the redirection result in an SSL verification error. Thus, one is unable to use `make_gitlab_blueprint` and must instead use the "raw" `Oauth2CustomerBlueprint` in order to pass `token_url_params=dict(verify=False)`.

This change is to add the ability to forward kwargs from `make_gitlab_blueprint` into `Oauth2CustomerBlueprint` for more customisationof functionality.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 * Dropped support for Python 2 and Python 3.5
 * If you are using the SQLAlchemy token storage, this project now depends on
   SQLAlchemy version 1.3.11 and above. `sqlalchemy-utils` is no longer necessary.
+* Add disabling of TLS certificate verification for Gitlab blueprint creation
 
 `3.3.1`_ (2021-03-01)
 ---------------------
--- a/flask_dance/contrib/gitlab.py
+++ b/flask_dance/contrib/gitlab.py
@@ -51,7 +51,7 @@ def make_gitlab_blueprint(
             specify the hostname, default is ``gitlab.com``
         verify_tls_certificates (bool, optional): Specify whether TLS
             certificates should be verified. Set this to ``False`` if
-            certificates fail to validate for self-hosted gitlab instances.
+            certificates fail to validate for self-hosted GitLab instances.
 
     :rtype: :class:`~flask_dance.consumer.OAuth2ConsumerBlueprint`
     :returns: A :ref:`blueprint <flask:blueprints>` to attach to your Flask app.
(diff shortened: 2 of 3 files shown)


singingwolfboy:
I would rather that we solve specific problems in these pre-set configurations, rather than operating as a catch-all where the developer has to know all the details of the `Oauth2CustomerBlueprint` in order to understand how to call `make_gitlab_blueprint()`. If your specific problem is "I need to turn off TLS verification because my GitLab server doesn't have a verifiable TLS certificate", then we can work with that!

Rather than a catch-all `**kwargs` arg, I suggest that you add an argument to `make_gitlab_blueprint()` named something like `verify`. We can then pass that argument down to all the places that it might be needed, including the `token_url_params`. We do a similar kind of thing in the Google configuration, with the arguments `offline`, `reprompt_consent`, and `hosted_domain`.

Please make sure to update the docstring for these arguments, and add an entry in the changelog as well.

willpwa:
Is there a way to add a word to that spellchecker's dictionary? It is complaining that `Gitlab` is not a word.

singingwolfboy:
> Is there a way to add a word to that spellchecker's dictionary? It is complaining that `Gitlab` is not a word.

Yep! Add it to `docs/spelling_wordlist.txt`

singingwolfboy:
Oh, one more thing I forgot -- please also add an automated test for this. You'll probably want to use the `responses` library to mock out the actual HTTP request, and then assert that `verify=False` was passed to it. Once we've got that, I think this will be ready to merge!

willpwa:
Unfortuantely, as far as I can tell, the `Responses` library can only check the returned object from a `requests` call. We instead want to check the parameters and this requires mocking.

I've got a test made but it doesn't test the actual `request` method, just that the variable that gets passed into the blueprint. If the test were to be improved, `requests.sessions.Session.request` would need to be mocked so that its parameters can be asserted.

The problem is that I don't know the flask-dance library well enough to be able to get that part of the code to execute in a testing environment and as such, have been unable to add it.

singingwolfboy:
Close enough! I'll merge this as-is, and see if I can fiddle with the test a bit to make it better. Thanks a lot!

Issue #357, opened 2021-04-24
Support for multi-tenancy based on host_name


spacebaire:
Hi

Is there any way that flask-dance supports multi-tenancy based on the host name?

I am in the process of adding multi-tenancy to my app (similar to https://stackoverflow.com/questions/51883773/flask-app-that-routes-to-multiple-top-level-and-sub-domains).  I have set host_name=True on the Flask app, and use a converter on the host parameter of my routes, which determine which tenant is being accessed (via the url_value_preprocessor decorator). This gets stored in g.host_entity. This works for my other routes when navigating to various URLs.

I understand that I need to set up multiple URLs on the google credentials dashboard.

When I look at app.url_map, I can see the google related route (login/google/authorized and login/google), but it does not have the host parameter as the other routes in the map has.

Is this supported somehow, or is this in the roadmap?

Do you believe this cannot be supported within flask-dance?  If it can, do you have some direction I can try to implement this myself maybe?

Thanks in advance,
Baire

singingwolfboy:
@spacebaire: I don't know anything about multi-tenancy in Flask apps, so I don't think I'll be able support this for you. However, I know a great deal about the architecture of Flask Dance, and you probably know a fair amount about multi-tenancy in Flask apps, so maybe we can put our heads together and figure something out?

Would you like to find a time to videochat and discuss your situation? I'd love it if Flask-Dance supported multi-tenancy, and I'd be happy to help you out if you could submit a pull request to make it more generally possible for others in the future.

spacebaire:
Hi

I played around with a local copy and eventually found that adding the "host" parameter to the add_url_rule calls in BaseOAuthConsumerBlueprint does the trick.  I had to set up several hosts on the google credentials dashboard I was testing my app with.  I was however able to authenticate from any of the domains I had set up, which was the goal.

To add the "host" parameter, I think it is best to open this up a bit broader and allow for adding any parameter.  I couldn't use **kwargs as it was already being used for something else, and I was worried there might be conflicts, so in the end I propose that we add a rule_kwargs (dict) on the make_*_blueprint calls, which then gets passed down (via OAuth1ConsumerBlueprint and OAuth2ConsumerBlueprint), until we can use them.  If they have kept their initial None value, then they are effectively ignored.  If the dict contains a value, then it is passed into both the login and authorized rules.

There are three potential concerns with my solution:

[...]

PR #358, opened 2021-04-26
#357 Add the possibility to add kwargs such as "host" from make*blueprint calls

spacebaire:
Having a multi-tenant site (based on the host name) requires that you use the "host" on the rule (by passing the "host" parameter to the add_url_rule call).  Secondly your Flask app needs to set host_matching to True and set static_host to the "main" host.  There are additional tricks around this, but the answer at https://stackoverflow.com/questions/51883773/flask-app-that-routes-to-multiple-top-level-and-sub-domains given a good indication of what is required.

Currently in flask-dance these rules are created from within BaseOAuthConsumerBlueprint in a rather hard-coded manner with no possibility to add parameters such as "host".

To pass in additional parameters, each make_*_blueprint is now able to take additional parameters (in the form of the rule_kwargs dict) and pass them down (via OAuth2ConsumerBlueprint or OAuth1ConsumerBlueprint).  Ultimately BaseOAuthConsumerBlueprint takes the additional parameter and adds them, defaulting to not adding any parameters if none were provided.

Unittests for Jira (and OAuth1) and Google (and OAuth2) and have been added to ensure the host can be set correctly.  

Parameters have been documented.

### DIFF
--- a/tests/consumer/test_oauth1.py
+++ b/tests/consumer/test_oauth1.py
@@ -574,5 +574,6 @@ def test_rule_kwargs():
     app = flask.Flask(__name__)
     app.secret_key = "secret"
     app.register_blueprint(blueprint, url_prefix="/login")
-    rules = [rule for rule in app.url_map.iter_rules() if rule.endpoint.startswith("google.")]
+    rules = [rule for rule in app.url_map.iter_rules() if rule.endpoint.startswith("test-service.")]
     assert all(rule.host == "example2.com" for rule in rules)
+    assert len(rules) == 2
--- a/tests/consumer/test_oauth1.py
+++ b/tests/consumer/test_oauth1.py
@@ -559,3 +559,23 @@ def test_custom_session_class():
     )
     assert isinstance(bp.session, CustomOAuth1Session)
     assert bp.session.my_attr == "foobar"
+
+
+def test_rule_kwargs():
+    blueprint = OAuth1ConsumerBlueprint(
+        "test-service",
+        __name__,
+        client_key="client_key",
+        client_secret="client_secret",
+        base_url="https://example.com",
+        request_token_url="https://example.com/oauth/request_token",
+        access_token_url="https://example.com/oauth/access_token",
+        authorization_url="https://example.com/oauth/authorize",
+        redirect_to="my_view",
+        rule_kwargs={"host":"example2.com"},
+    )
+    app = flask.Flask(__name__)
+    app.secret_key = "secret"
+    app.register_blueprint(blueprint, url_prefix="/login")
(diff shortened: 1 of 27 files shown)


tests/consumer/test_oauth1.py
```diff
@@ -556,3 +556,23 @@ def test_custom_session_class():
...
+        request_token_url="https://example.com/oauth/request_token",
+        access_token_url="https://example.com/oauth/access_token",
+        authorization_url="https://example.com/oauth/authorize",
+        redirect_to="my_view",
+        rule_kwargs={"host":"example2.com"},
+    )
+    app = flask.Flask(__name__)
+    app.secret_key = "secret"
+    app.register_blueprint(blueprint, url_prefix="/login")
+    rules = [rule for rule in app.url_map.iter_rules() if rule.endpoint.startswith("google.")]
```
singingwolfboy, line 577: Please assert that there should be exactly two rules that match. Right now, I believe this test is only passing because _zero_ rules match this filter.

tests/consumer/test_oauth2.py
```diff
@@ -677,3 +677,23 @@ def test_custom_session_class():
...
+        base_url="https://example.com",
+        authorization_url="https://example.com/oauth/authorize",
+        token_url="https://example.com/oauth/access_token",
+        redirect_url="http://mysite.cool/whoa?query=basketball",
+        rule_kwargs={"host":"example2.com"},
+    )
+    app = flask.Flask(__name__)
+    app.secret_key = "secret"
+    app.register_blueprint(blueprint, url_prefix="/login")
+    rules = [rule for rule in app.url_map.iter_rules() if rule.endpoint.startswith("google.")]
```
singingwolfboy, line 698: Same with asserting on the number of matching rules.

tests/consumer/test_oauth1.py
spacebaire: Done.  I've done the same for the oauth2, google and jira tests.  The jira test still said "google", so had the same issue.

spacebaire:
I've made the pytest corrections you have suggested.  I've also fixed the typo in the docstring, which was causing the sphinx build to fail.

spacebaire:
@singingwolfboy , I've made the corrections for black as well (sorry, I missed that requirement initially).  Is anything else expected from me on this pull request?  (this is my first pull request ever)  Baire

singingwolfboy:
No, I think this is great! I'm sorry I haven't been particularly communicative here -- I don't actively use Flask-Dance in any of my projects these days, but I'm glad that others do, and are actively contributing new features!

Thank you for all the work you've put into this, including writing tests and docs. I really appreciate it!

PR #360, opened 2021-05-05
use keyword-only arguments

singingwolfboy:
See https://www.python.org/dev/peps/pep-3102/. I intended to use this feature when I first created this library, but Python 2.7 compatibility prevented me from doing so, and I forgot about it until recently. Since this has the potential of causing backwards-incompatibility, we will need to bump the major version number.

@daenney, do you have any thoughts about this change?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ Changelog
   pre-set configurations have been updated to also accept a `rule_kwargs` parameter
   as well.
 * Minimum supported version of Flask is now 1.0.4.
+* The blueprint classes and the pre-set configurations now use keyword-only arguments,
+  as defined in `PEP-3102`_.
 
 `4.0.0`_ (2021-04-10)
 ---------------------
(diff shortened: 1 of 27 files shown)


daenney:
I like this. I've always found the behaviour where positional args can fill in keyword args confusing. It's almost always an error, and I like the self-documenting nature of keyword args.

One question though, in your changeset you're keeping `client_id` and `client_secret` as old-style arguments, kwarg but could be filled by positional. Why is that?

singingwolfboy:
> One question though, in your changeset you're keeping `client_id` and `client_secret` as old-style arguments, kwarg but could be filled by positional. Why is that?

I did that because `client_id` and `client_secret` are necessary for Flask-Dance to work, and so they are almost like positional, required arguments. I intentionally kept them as the first two arguments in every pre-set configuration, for that reason. The only time that it's OK to _not_ specify those arguments is when you are loading those values in the `flask.config` object instead, which is why they are not _actually_ positional, required arguments. You'll also notice that the automated tests use this pattern a fair amount, specifying the `client_id` and `client_secret` as positional arguments. I want that to be a supported pattern, since it seems fairly clear to me.

Does that make sense, and do you agree?

daenney:
Yup, that makes sense!

daenney (approved):

PR #361, opened 2021-05-05
install specific Flask version for tests

singingwolfboy:
We should verify that we _actually_ support the lowest version of Flask that we claim to support. That means running the automated tests with that version.

I discovered that the latest version of `pytest` doesn't run with Flask 0.7 anymore, which is why I'm bumping up the minimum supported version to Flask 1.0.4. I believe this is due to [this Flask pull request](https://github.com/pallets/flask/issues/3275).

Since we are changing the minimum version of Flask that we support, this will require a major version bump.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install specific Flask version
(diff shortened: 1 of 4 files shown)


daenney (approved):
This makes sense to me. It's too bad we can't pin flask on `>=1.0.0` but it is what it is.

PR #371, opened 2021-10-13
added fitbit configuration

lila:
Thanks for maintaining this project.  I've added fitbit to the list of supported oauth2 providers.  I used the other providers as a template and have tested this with https://github.com/lila/flask-dance-fitbit.  it works but there are probably some improvements that are possible.  
  
  - the fitbit api uses a base64 encoded string "client_id:client_secret" as an authorization header, it would be nice to store this in the blueprint, but not sure how to do that.
  -  the token renewal uses the same token api endpoint, but requires some specific parameters (`grant_type`, `refresh_token`, and a custom authorization header).  I'm not sure how standard this is and if the underlying libraries do the right thing.  upshot is that the client probably has to do this.  

thanks again, i hope this is a helpful contribution to others...

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -54,6 +54,7 @@ Strava
 Twitter
 Zoho
 Atlassian
+Fitbit
 
 Pytest
 Werkzeug
--- a/tests/contrib/test_fitbit.py
+++ b/tests/contrib/test_fitbit.py
@@ -27,7 +27,7 @@ def test_blueprint_factory():
     )
     assert isinstance(fitbit_bp, OAuth2ConsumerBlueprint)
     assert fitbit_bp.session.scope == "user:email"
-    assert fitbit_bp.session.base_url == "https://www.fitbit.com/"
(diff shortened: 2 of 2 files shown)


singingwolfboy:
Hi @lila, this is great! Thank you so much for the pull request 😄 

Looks like the tests are failing due to an unrelated environmental issue. I've already merged a fix in #372, so if you rebase your pull request onto the latest version of `main`, the tests should pass! Please do so -- I need the tests to all be passing before I can merge this pull request.

Also, it would be fantastic if you want to add https://github.com/lila/flask-dance-fitbit to the documentation! To do so, edit the file `docs/quickstart.rst` -- it should be pretty straightforward to add it. All of the other links in that file are to repos that I own, but that doesn't need to be the case; I would actually prefer to have more people involved! So there's no need to "contribute" your repository, you can keep ownership of it yourself, and just add the link to the docs.

singingwolfboy:
Looks like your tests are failing. You'll also need to tell Sphinx that "Fitbit" is a word, by adding it to `docs/spelling_wordlist.txt`.

lila:
great, thank you!

PR #394, opened 2022-06-05
Add OpenStreetMap (OSM) provider

mvexel:
This adds OpenStreetMap as an OAuth2 provider.

I accidentally created (and then deleted in a later commit) a test application in the root directory. This application contained a client secret, but I deleted the corresponding OAuth2 application on the OpenStreetMap side, so this should not be a security issue.

The only quirk with OSM is that the `base_url` has a path suffix. This means that requests need to be composed without a leading slash, which may be confusing to users.


singingwolfboy:
Hi @mvexel, I'm sorry I missed this pull request! I just released version 6.1.0, and I cleaned up this pull request and included it as https://github.com/singingwolfboy/flask-dance/commit/b8430a50aaccba6c361fc88043b44b2b58bde1de. As a result, I'm closing this pull request, since your code has effectively already been merged! Thanks for your contribution ✨

Issue #396, opened 2022-08-02
flask-dance 6.0.0 incompatable with Flask >=2.2.0


rez10191:
Hello,

Flask 2.2.0 was released today.  This release included this change list: https://github.com/pallets/flask/commit/0b2f809f9b56861dae3ae5154d73b4afaf632a0a which removes the `_lookup_app_object` function that is imported from `flask_dance/contrib/google.py`, as well as many other `contrib` modules by the looks of it.  The function was properly marked as a private.

Thanks for your great work on this project!

Richard Zuber

BlitzJB:
Faced the same issue! @rez10191 Can you point me in the general direction of a workaround? I could put some time into this and submit a PR!

singingwolfboy:
I just released Flask-Dance 6.1.0, which switches to using `flask.g` for storing information, as recommended by the Flask release notes. Thanks for the heads-up about this change!

PR #397, opened 2022-08-03
Made fixes to extend compatibility untill Flask 2.3.0

BlitzJB:
Due to changes from `flask 2.2.0`, `_lookup_app_object` which is commonly used across all contrib modules is now depreciated. So i mimicked `_lookup_app_object` from `flask 2.1.3` in `flask_dance.utils` and changed imports in all contrib modules to import `_lookup_app_object` from `flask_dance.utils` instead

### DIFF
--- a/flask_dance/contrib/atlassian.py
+++ b/flask_dance/contrib/atlassian.py
@@ -1,9 +1,10 @@
 from functools import partial
 
 from flask import _app_ctx_stack as stack
-from flask.globals import LocalProxy, _lookup_app_object
+from flask.globals import LocalProxy
 
 from flask_dance.consumer import OAuth2ConsumerBlueprint
+from flask_dance.utils import _lookup_app_object
 
 __maintainer__ = "Przemyslaw Kanach <kanach16@gmail.com>"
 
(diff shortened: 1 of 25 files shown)


BlitzJB:
To comment on the failing checks, they are all depreciation warnings on `_app_ctx_stack` which i imported from `flask.global` to use in `flask_dance.utils._lookup_app_object`, I couldn't think of a more graceful way to achieve the same. If there seem to be, please do let me know and I'll make the changes

BlitzJB:
This was in response to #396

singingwolfboy:
Thanks for sending this pull request, but I ended up solving this problem a different way. [According to the release notes for Flask 2.2.0](https://github.com/pallets/flask/blob/main/CHANGES.rst#version-220), we can and should store information on the `flask.g` object instead of the `_app_ctx_stack` object. This also makes the change backwards-compatible for all versions of Flask that Flask-Dance supports.

I implemented that change in #398, so I'm closing this pull request. However, if your goal is compatibility with the latest version of Flask, try installing Flask-Dance 6.1.0, which I just released today with this change. If you have problems, let me know!

PR #398, opened 2022-08-04
Switch from _app_ctx_stack to g

singingwolfboy:
This is to support Flask 2.2.0; see #396.


daenney (approved):
Code wise the changes look fine to me, though some tests are still rather upset.

On the Sphinx one, it fails because Blinker has moved to `https://blinker.readthedocs.io`, so it's just a matter of updating that link.

PR #403, opened 2022-10-03
Add ORCID OAuth2 provider

ml-evs:
This PR adds support for ORCID (https://orcid.org/) and its sandbox server (https://sandbox.orcid.org) via a blueprint init option `use_orcid_sandbox`.

It is slightly tricky to test this as OAuth configs at ORCID require human intervention at ORCID's end, and even then, the ORCID sandbox does not allow non-HTTPS callback URIs (like localhost). I was able to use this code to authorize via the ORCID sandbox by monkey-patching the authorized URL in the base blueprint to point at [Google's OAuth playground](https://developers.google.com/oauthplayground/)  using the `openid` scope (might be useful if this was easier to do).

Thanks for flask-dance!

### DIFF
--- a/flask_dance/contrib/orcid.py
+++ b/flask_dance/contrib/orcid.py
@@ -1,0 +1,93 @@
+from flask import g
+from werkzeug.local import LocalProxy
+
+from flask_dance.consumer import OAuth2ConsumerBlueprint
+
+__maintainer__ = "Matthew Evans <git@ml-evs.science>"
+
+
+def make_orcid_blueprint(
+    client_id=None,
+    client_secret=None,
+    *,
+    scope=None,
+    redirect_url=None,
+    redirect_to=None,
+    login_url=None,
+    authorized_url=None,
+    session_class=None,
+    storage=None,
+    rule_kwargs=None,
+    use_orcid_sandbox=False,
+):
+    """
+    Make a blueprint for authenticating with ORCID (https://orcid.org)
+    using OAuth2.
+
+    This requires a client ID and client secret from ORCID. You should
+    either pass them to this constructor, or make sure that your Flask
+    application config defines them, using the variables
+    :envvar:`ORCID_OAUTH_CLIENT_ID` and :envvar:`ORCID_OAUTH_CLIENT_SECRET`.
+
+    The ORCID Sandbox API (https://sandbox.orcid.org) will be used if
+    the ``use_orcid_sandbox`` argument is set to true.
+
+    Args:
+        client_id (str): The client ID for your application on ORCID.
+        client_secret (str): The client secret for your application on ORCID
+        scope (str, optional): comma-separated list of scopes for the OAuth token
+        redirect_url (str): the URL to redirect to after the authentication
+            dance is complete
+        redirect_to (str): if ``redirect_url`` is not defined, the name of the
+            view to redirect to after the authentication dance is complete.
+            The actual URL will be determined by :func:`flask.url_for`
+        login_url (str, optional): the URL path for the ``login`` view.
+            Defaults to ``/orcid``
+        authorized_url (str, optional): the URL path for the ``authorized`` view.
+            Defaults to ``/orcid/authorized``.
+        session_class (class, optional): The class to use for creating a
+            Requests session. Defaults to
+            :class:`~flask_dance.consumer.requests.OAuth2Session`.
+        storage: A token storage class, or an instance of a token storage
+                class, to use for this blueprint. Defaults to
+                :class:`~flask_dance.consumer.storage.session.SessionStorage`.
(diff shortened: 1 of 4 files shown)


flask_dance/contrib/orcid.py
```diff
@@ -0,0 +1,93 @@
...
+
+    orcid_bp = OAuth2ConsumerBlueprint(
+        "orcid",
+        __name__,
+        client_id=client_id,
+        client_secret=client_secret,
+        scope=scope,
+        base_url=base_url,
+        authorization_url="https://orcid.org/oauth/authorize",
+        token_url="https://orcid.org//oauth/token",
```
singingwolfboy, line 74: Did you intend for this to have two slashes before `oauth`?

tests/contrib/test_orcid.py
```diff
@@ -0,0 +1,101 @@
...
+    assert isinstance(orcid_bp, OAuth2ConsumerBlueprint)
+    assert orcid_bp.session.scope == "user:email"
+    assert orcid_bp.session.base_url == "https://api.orcid.org"
+    assert orcid_bp.session.client_id == "foo"
+    assert orcid_bp.client_secret == "bar"
+    assert orcid_bp.authorization_url == "https://orcid.org/oauth2/authorize"
+    assert orcid_bp.token_url == "https://orcid.org/oauth2/token"
+
+
+def test_blueprint_factory():
```
singingwolfboy, line 38:
```suggestion
def test_blueprint_factory_for_sandbox():
```

singingwolfboy (changes requested):
This is a great start; thanks for making this pull request! However, your tests are failing, and you still need to edit the [`docs/providers.rst`](https://github.com/singingwolfboy/flask-dance/blob/main/docs/providers.rst) file so that this shows up in the documentation. You also need to update the changelog. See the [README in the `contrib` dir](https://github.com/singingwolfboy/flask-dance/blob/main/flask_dance/contrib/README.rst).

ml-evs:
Thanks for the review, rushed this out a bit and should have left it as a draft until I could finish it off. I should have addressed all your concerns (apologies if you had to manually authorize rerun the workflows!)

flask_dance/contrib/orcid.py
```diff
@@ -0,0 +1,97 @@
...
+    Make a blueprint for authenticating with ORCID (https://orcid.org)
+    using OAuth2.
+
+    This requires a client ID and client secret from ORCID. You should
+    either pass them to this constructor, or make sure that your Flask
+    application config defines them, using the variables
+    :envvar:`ORCID_OAUTH_CLIENT_ID` and :envvar:`ORCID_OAUTH_CLIENT_SECRET`.
+
+    The ORCID Sandbox API (https://sandbox.orcid.org) will be used if
+    the ``use_orcid_sandbox`` argument is set to true.
```
singingwolfboy, line 33: I actually think that `sandbox` is a clear enough name. The function is already named `make_orcid_blueprint()`, including "orcid" in the argument name seems unnecessary. Let's keep it concise!

ml-evs: Fair, it felt a bit ambiguous to me (given that its really just contacting a different OAuth server) but in context it should be clear enough!

singingwolfboy:
Looks great, thank you! I'll cut a new release in the next few days.

ml-evs:
> Looks great, thank you! I'll cut a new release in the next few days.

Thanks, I appreciate it!

singingwolfboy:
@ml-evs I just published version 6.2.0 a few hours ago, including your change! Thanks again for contributing to this project. 😄

ml-evs:
Awesome, thanks for letting me know.

mpound:
@ml-evs Do you have a working example using orcid for authentication you could share?  similar to the github example on https://flask-dance.readthedocs.io/en/latest/ ? 
I tried the obvious substitutions of 'orcid' for 'github' in the python script, but get weird errors like "Incorrect OAuth Link for client id , missing scope parameter."  However changing the call to explicitily set the scope: make_orcid_blueprint(scope=["authenticate"])   results in a different error "CSRF Warning! State not equal in request and response."     A working example might help clarify what I am doing wrong.
thanks.

ml-evs:
Hi @mpound, sure thing, I can type something up when I get home. Have you registered an app with the sandbox ORCID API already? For me it was process with manual approval from the ORCID team. 

Just in case it is a simple issue, I can see that in my code (unfortunately closed source for now) my scope is set to `"/authenticate"` rather than the list `["authenticate"]`.

mpound:
ok, I got a little farther with your hints.   It appears to authenticate correctly at
    resp = orcid.get("/user")
    assert resp.ok

But resp.ok is False.  I did register the test app with sandbox orcid on the developer tools section of my account, which gave me the CLIENT_ID and CLIENT_SECRET.

PR #405, opened 2022-10-12
add @daenney as official maintainer

singingwolfboy:
Hey @daenney, I should have done this awhile ago, but better late than never, I suppose! Would you like to be in the `pyproject.toml` file as a maintainer for the project? And if so, which email would you like to use? (It will be public!)


singingwolfboy:
Nope, looks like both `name` and `email` are optional keys here! I've taken out the `email` key, so it's just your name. Is that what you want?

daenney:
Yap, that'll do! Thanks

daenney (approved):

Issue #293, opened 2019-12-20
Support for PKCE?


xscheck:
I have a working version of a Flask app using FD that authenticates my REST APIs successfully when the APIs are accessed from a browser.  I am using Google OAuth2. Much appreciate what FD brings to the table. 

I am trying to do the same now with a mobile app: it uses a subset of the REST APIs but user needs to be authenticated first. I believe I have to have PKCE flow, not Implicit flow. (I want the API server to get the auth token, not the user's device - the app on the device should not have the Secrets)

I've googled, checked the FD docs and the open-closed issues here but could not find anything specific on the topic. Closest is issue #265. Is that same as PKCE support (or isn't that what was meant there? Yes, I am second guessing)?  I don't see code-verifier/code-challenge and code-challenge-method in the Providers' doc.

OR, there is some other way to manage the authentication from FD client to a mobile app? 

Sorry if newbie question. I am also trying to implement (for example using AppAuth on Android) but am not seeing all the pieces (yet) to make this work.

Thanks in advance for any insights.

daenney:
#265 is not PKCE. Authorization Code Grant is the most common flow for oauth2, so yes we support that. PKCE is an extension: https://oauth.net/2/pkce/

> PKCE (RFC 7636) is an extension to the Authorization Code flow to prevent certain attacks and to be able to securely perform the OAuth exchange from public clients.
> It is primarily used by mobile and JavaScript apps, but the technique can be applied to any client as well.

Flask-Dance doesn't do much with oauth itself, we use requests-oauthlib and thus oauthlib instead which does support PKCE. The problem is we currently don't expose that ability in FD, and we'd have to override https://github.com/oauthlib/oauthlib/blob/794c9fb1305141903f4cd182d7d0aab959e86219/oauthlib/oauth2/rfc6749/request_validator.py#L579 by construction our own `RequestValidator`. It doesn't look like requests-oauthlib currenctly allows for that, so we'd have to start with a change in that library first.

@singingwolfboy Do you have a sense of what we would have to do here? I think supporting PKCE makes a lot of sense and is something I'd like to see us have.

singingwolfboy:
It would be nice to support PKCE. It would be nice to support OIDC. It would be nice to support a _lot_ of things: OAuth is a very, very broad standard that has a lot of different optional pieces.

[...]

PR #406, opened 2022-10-19
Add support for authorization flow with PKCE 

mukiblejlok:
Hello @singingwolfboy,

I know there is already issue #293 with a discussion about supporting PKCE flow. Since it's true that PKCE is mostly used with mobile apps, there are some OAuth2 servers (for example the one that we use in our company) that require clients to use PKCE flow. As the [link](https://oauth.net/2/pkce/) says:

> PKCE was originally designed to protect the authorization code flow in mobile apps, but its ability to prevent authorization code injection makes it useful for every type of OAuth client, even web apps that use a client secret.

 As far as I'm aware there is no Open Source package for Flask that supports it.

I took some initial effort to implement the basic support for it in this PR. Being aware I'm far away from done (tests and docs updates are not there yet), I wanted to start the conversation and ask if you see the point in adding the PKCE support to Flask-Dance in the form I have proposed. 

If so, I would be happy to continue my work on that pull request, especially since it would be my first bigger Open Source contribution. Something that I wanted to start doing for a while.


singingwolfboy:
Thank you for doing this, and I'm sorry it took me a few days to review this!

I like your initiative here, but I think it would be better to implement this in a different way. [OAuthlib has supported PKCE](https://oauthlib.readthedocs.io/en/latest/feature_matrix.html?highlight=pkce#oauth-2-0) since version 3.0, [according to their changelog](https://github.com/oauthlib/oauthlib/blob/master/CHANGELOG.rst#300-2019-01-01). Rather than implementing these functions in a `flask_dance/consumer/pkce.py` file, can we use the right function arguments to make OAuthlib do this verifications?

Check the docs for [`prepare_request_uri`](https://oauthlib.readthedocs.io/en/latest/oauth2/clients/webapplicationclient.html#oauthlib.oauth2.WebApplicationClient.prepare_request_uri) -- looks like we can just pass `code_challenge` and `code_challenge_method` in the kwargs to make this work. I think we can do something similar with the [`prepare_request_body`](https://oauthlib.readthedocs.io/en/latest/oauth2/clients/webapplicationclient.html#oauthlib.oauth2.WebApplicationClient.prepare_request_body) function -- pass the `code_verifier` in the kwargs. Can you see how to do that? I don't think that requests-oauthlib has any special handling for PKCE, but I don't think it will get in the way, either!

mukiblejlok:
Thank you for the very detailed feedback. I agree, it would be much better to use the features of `OAuthlib` instead of writing your own as I did. I was not aware of the PKCE support in `OAuthlib 3`.  

As for passing the PCKE parameters to the `prepare_request_uri` and `prepare_request_body`, I believe that is what I am doing anyway, but not explicitly.  `authorization_url` method from `request_oautlib` passes kwargs to `_client.prepare_request_uri` and `fetch_token`  passes kwargs to `_client.prepare_request_body`. 

Still, the core challenge in PKCE flow is to generate a `code_verifier` and `code_challenge` pair at the login stage, then save the `code_verifier` challenge in the context of the current OAuth Session (by storing it with the `state` param) and then to use the `code_verifier` at the authorization stage. In my opinion that would have to be implemented in the `login` and `authorized` methods of `OAuth2ConsumerBlueprint`.  But if you have any other ideas, I'd be happy to do it your way.

flask_dance/consumer/oauth2.py
```diff
@@ -152,6 +161,8 @@ def __init__(
         self.token_url_params = token_url_params or {}
         self.redirect_url = redirect_url
         self.redirect_to = redirect_to
+        self.code_challenge_method = code_challenge_method
+        self._use_pkce = is_valid_code_challenge_method(self.code_challenge_method)
```
singingwolfboy, line 165: I don't think we need this. I'd say there are two good ways to structure this API:

1. This `__init__` method accepts both a `use_pkce` boolean (default to False) _and_ a `code_challenge_method` string (default to "S256"). If `use_pkce` is True, then we do PKCE. Otherwise, we don't.
2. This `__init__` method accepts a `code_challenge_method` string (default to None). If it is set to a truthy string, then we do PKCE with the challenge method provided. If it is set to a falsy value, including the default None, then we don't do PKCE at all.

Personally, I think that option 1 is the better one. The spec says that "plain" should be the default, but it also says that devices that _can_ do sha256 hashing MUST do so, so "S256" seems like a reasonable default for the method in this case. But if you prefer option 2, let's discuss it!

Either way, I don't think we need to worry about validating the value of `code_challenge_method`. The spec says that its optional, and [the oauthlib implementation](https://github.com/oauthlib/oauthlib/blob/0adbbe10ed8ef822d1c780987fffc56670ce3f9f/oauthlib/oauth2/rfc6749/clients/base.py#L504) falls back on "plain" if the `code_challenge_method` argument is anything other than "S256", so I don't think we need to raise an exception if it's something unexpected.

flask_dance/consumer/oauth2.py
```diff
@@ -203,12 +214,25 @@ def teardown_session(self, exception=None):
...
+            })
+
         url, state = self.session.authorization_url(
             self.authorization_url, state=self.state, **self.authorization_url_params
         )
         state_key = f"{self.name}_oauth_state"
         flask.session[state_key] = state
         log.debug("state = %s", state)
+        if code_verifier:
+            code_verifier_key = f"{self.name}_{state}_oauth_code_verifier"
```
singingwolfboy, line 233: I don't think we need to include the `state` token in this key name. I would suggest:
```suggestion
            code_verifier_key = f"{self.name}_pkce_code_verifier"
```

flask_dance/consumer/oauth2.py
```diff
@@ -254,6 +278,18 @@ def authorized(self):
...
+            if code_verifier_key not in flask.session:
+                # can't find code_verifier, so redirect back to login view
+                log.info("code_verifier not found, redirecting user to login")
+                return redirect(url_for(".login"))
+
+            code_verifier = flask.session[code_verifier_key]
+            log.debug("code_verifier = %s", code_verifier)
+            del flask.session[code_verifier_key]
+            self.token_url_params.update({"code_verifier": code_verifier})
```
singingwolfboy, line 291: You don't need to use `.update()` with only one value -- it's needlessly complex.
```suggestion
            self.token_url_params["code_verifier"] = code_verifier
```

flask_dance/consumer/pkce.py
```diff
@@ -0,0 +1,42 @@
+import hashlib
```
singingwolfboy, line 1: I think this whole file can be deleted, in favor of the functions provided by `oauthlib`.

tests/consumer/test_oauth2.py
```diff
@@ -90,6 +91,43 @@ def test_login_url():
     assert location.query_dict["state"] == "random-string"
 
 
+@responses.activate
+def test_login_url_with_pkce():
```
singingwolfboy, line 95: Thanks for adding these tests! We also need to test the `authorization_url` as well, though. Check the test coverage to be sure you've added tests to cover all the lines you added.

singingwolfboy (changes requested):
> Thank you for the very detailed feedback. I agree, it would be much better to use the features of `OAuthlib` instead of writing your own as I did. I was not aware of the PKCE support in `OAuthlib 3`.

Glad I could help!
> 
> As for passing the PCKE parameters to the `prepare_request_uri` and `prepare_request_body`, I believe that is what I am doing anyway, but not explicitly. `authorization_url` method from `request_oautlib` passes kwargs to `_client.prepare_request_uri` and `fetch_token` passes kwargs to `_client.prepare_request_body`.

I just took another look at the code, and you are absolutely correct! Sorry I missed that.
 
> Still, the core challenge in PKCE flow is to generate a `code_verifier` and `code_challenge` pair at the login stage, then save the `code_verifier` challenge in the context of the current OAuth Session (by storing it with the `state` param) and then to use the `code_verifier` at the authorization stage. In my opinion that would have to be implemented in the `login` and `authorized` methods of `OAuth2ConsumerBlueprint`. But if you have any other ideas, I'd be happy to do it your way.

I went through your code and added a few comments. I think we can simplify things a bit, and you need tests for the authorization URL, but this is actually closer to merge-ready than I thought!

flask_dance/consumer/oauth2.py
mukiblejlok: Here I was not sure what would be the preferred interface for the method. But option number 1 looks good to me, I will update it this way.

flask_dance/consumer/oauth2.py
mukiblejlok: I was afraid of an edge case where more than one authentication process happens at once. 
I don't know if it's even a problem, since if the process is interrupted at the login stage and started again we will update the same key with the new `code_verifier` value which is the desired behavior.
I went for saving the `code_verifier` under the state aware key, but if you think that's an excessive solution, I don't have anything against simplifying it.

flask_dance/consumer/oauth2.py
mukiblejlok: I agree, I will simplify it.

tests/consumer/test_oauth2.py
mukiblejlok: Yes, I was aware those tests were missing. I will add them for sure.

flask_dance/consumer/oauth2.py
```diff
@@ -114,6 +120,9 @@ def __init__(
                 :class:`~flask_dance.consumer.storage.session.SessionStorage`.
             rule_kwargs (dict, optional): Additional arguments that should be passed when adding
                 the login and authorized routes. Defaults to ``None``.
+            code_challenge_method: Code challenge method to be used in authorization code flow with PKCE
+                instead of client secret. Currently, only ``S256`` is supported. For more details please refer to TODO.
```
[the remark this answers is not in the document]
mukiblejlok: My idea was to add a reference either to [RFC7636](https://www.rfc-editor.org/rfc/rfc7636#section-4.2) or to the internal documentation section about PKCE in OAuth2.

This brings a question, do you want me to add a chapter about PKCE to the [How OAuth works](https://github.com/singingwolfboy/flask-dance/blob/main/docs/how-oauth-works.rst)?

I can do that, but please consider I am not a native English speaker and the text might need some polishing.

flask_dance/consumer/base.py
```diff
@@ -146,7 +146,7 @@ def token(self, value):
         if _token and _token.get("expires_in"):
             # Set the `expires_at` value, overwriting any value
             # that may already be there.
-            delta = timedelta(seconds=_token["expires_in"])
+            delta = timedelta(seconds=int(_token["expires_in"]))
```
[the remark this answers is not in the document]
mukiblejlok: You are right, this is unrelated to the core subject of this PR. 
But while trying to adopt Flask-Dance to the OAuth server I was dealing with I noticed it was returning the `expires_in` value as a string. AFAIK this is not according to OAuth2 specification, but on the other hand, it doesn't hurt the values that are already provided as ints, and it helps with tokens with too many strings attached (pun intended).

If you don't want to have it, I can remove it, and preprocess the token before saving it to the blueprint.
```python
@oauth_authorized.connect
def redirect_to_next_url(blueprint, token):
    token = preprocess_token(token)
    blueprint.token = token
```

flask_dance/consumer/oauth2.py
singingwolfboy: There should not be more than one authentication process happening simultaneously, certainly not for the same OAuth provider. I'm not even sure the `self.name` reference is needed in the key name (for this or the `state` token), although it may help for debugging; adding a `state` reference is certainly excessive.

singingwolfboy (changes requested):
This is fantastic! I think there are only three things left to do in this pull request:

* Remote the `state` reference from the `code_verifier_key` variable
* In `pyproject.toml`, change the `oauthlib` dependency to `oauthlib>3.0`
* Add a note to the "unreleased" section of `CHANGELOG.rst`, briefly mentioning this change

Thank you so much for contributing this feature! Once you've made those last three changes, and the tests have passed, I'll merge this pull request. I'll try to add something to the documentation about this as well, so I probably won't make a new release right away, but I will soon.

flask_dance/consumer/oauth2.py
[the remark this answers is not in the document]
singingwolfboy: I think a reference to the RFC is enough. But I'll try to add something to the documentation, as well! I'll make a separate pull request for that, and you can review it to be sure that it's accurate. 😄

mukiblejlok:
I am happy I could add a small contribution to your great package. Due to your support, it was a very pleasant experience. I will try to get involved more in the future.

PR #407, opened 2022-12-13
Adds base_url to allow different data center usage

ajay-k:
Gives the ability to change base_url, currently, it defaults to the US data center (https://api.nylas.com/) but customers in EU are unable to make requests since the EU data center is at (https://ireland.api.nylas.com)


mrashed-dev (approved):
Looks good to me!

flask_dance/contrib/nylas.py
```diff
@@ -49,6 +50,8 @@ def make_nylas_blueprint(
                 :class:`~flask_dance.consumer.storage.session.SessionStorage`.
         rule_kwargs (dict, optional): Additional arguments that should be passed when adding
             the login and authorized routes. Defaults to ``None``.
+        base_url (str,optional): A string that points to the data-center the application is stored on
+        Defaults to US Data-center(https://api.nylas.com/) EU Data-center (https://ireland.api.nylas.com)
```
daenney, line 54: This line needs proper indenting to render correctly in the docs.

flask_dance/contrib/nylas.py
```diff
@@ -59,9 +62,9 @@ def make_nylas_blueprint(
         client_id=client_id,
         client_secret=client_secret,
         scope=scope,
-        base_url="https://api.nylas.com/",
-        authorization_url="https://api.nylas.com/oauth/authorize",
-        token_url="https://api.nylas.com/oauth/token",
+        base_url=base_url,
+        authorization_url= base_url + "/oauth/authorize",
+        token_url= base_url + "/oauth/token",
```
daenney, line 67: We use f-strings for interpolation across this project, or `base_url.relative("oauth/authorize")` etc.

flask_dance/contrib/nylas.py
ajay-k: I made these changes, could you confirm if it looks good to you now?

daenney: Yup, that seems better.

daenney:
This looks mostly good, though we'd need a test in https://github.com/singingwolfboy/flask-dance/blob/main/tests/contrib/test_nylas.py too to ensure the `base_url` passes through correctly.

ajay-k:
> This looks mostly good, though we'd need a test in https://github.com/singingwolfboy/flask-dance/blob/main/tests/contrib/test_nylas.py too to ensure the `base_url` passes through correctly.
> 
> You can add something like `def test_blueprint_factory_base_url(make_app):`, call the constructor with a value in `base_url` and assert that the right URLs are found in the `base_url`, `token_url` etc.

@daenney Made these on this [PR](https://github.com/singingwolfboy/flask-dance/pull/408)

daenney:
> > This looks mostly good, though we'd need a test in https://github.com/singingwolfboy/flask-dance/blob/main/tests/contrib/test_nylas.py too to ensure the `base_url` passes through correctly.
> > You can add something like `def test_blueprint_factory_base_url(make_app):`, call the constructor with a value in `base_url` and assert that the right URLs are found in the `base_url`, `token_url` etc.
> 
> @daenney Made these on this [PR](https://github.com/singingwolfboy/flask-dance/pull/408)

Thanks, but we need the tests in this PR, not a separate one. Otherwise they won't run here and validate the changes.

ajay-k:
> > > This looks mostly good, though we'd need a test in https://github.com/singingwolfboy/flask-dance/blob/main/tests/contrib/test_nylas.py too to ensure the `base_url` passes through correctly.
> > > You can add something like `def test_blueprint_factory_base_url(make_app):`, call the constructor with a value in `base_url` and assert that the right URLs are found in the `base_url`, `token_url` etc.
> > 
> > 
> > @daenney Made these on this [PR](https://github.com/singingwolfboy/flask-dance/pull/408)
> 
> Thanks, but we need the tests in this PR, not a separate one. Apologies for not being clearer on that. Otherwise they won't run here and validate the changes.

Thanks for the help, I made the tests in this PR now

ajay-k:
@daenney When you get a chance, could you review the changes I made?

flask_dance/contrib/nylas.py
```diff
@@ -18,6 +18,7 @@ def make_nylas_blueprint(
     session_class=None,
     storage=None,
     rule_kwargs=None,
+    base_url="https://api.nylas.com/",
```
singingwolfboy, line 21: Rather than specifying a `base_url`, which may not point to Nylas at all, I would prefer if this function accepted a `subdomain` argument. The default value could be `None`, which sets the `base_url` to "https://api.nylas.com/". For this use-case, the developer could pass `subdomain="ireland"`, which would result in setting the `base_url` to "https://ireland.api.nylas.com/".

singingwolfboy (changes requested):

flask_dance/contrib/nylas.py
ajay-k: @singingwolfboy Thanks for the feedback, I've made the changes. Could you confirm if that's what you had in mind?

singingwolfboy: No, I intended something more like this:
```python
if subdomain:
    base_url = f"https://{subdomain}.api.nylas.com"
else:
    base_url = "https://api.nylas.com"
```

Also, please be sure that you run the tests locally when making changes to this pull request. This pull request will not be merged as long as there are any test failures.

singingwolfboy:
@ajay-k are you still working on this pull request?

singingwolfboy:
> @ajay-k are you still working on this pull request?

It's been two weeks, and @ajay-k has not responded. I will assume that the answer is no, and close this pull request.

PR #408, opened 2022-12-13
Update test_nylas.py

ajay-k:
Adding tests to ensure base_url parses correctly


daenney:
Thanks for this! But please add them as part of #407, not as a separate PR. Otherwise the tests don't run together.

Issue #286, opened 2019-09-30
Azure auth - token expired error.


kknd4eva:
Hi there. I've implemented flask dance on a python website running via AWS Elastic Beanstalk. Initial authentication works fine, however after a period of time the token expires, and I get the below errors (500 server error on the site). I would have expected an expired token would redirect the user to login again? 

```

[Mon Sep 30 01:16:56.209088 2019] [:error] [pid 2778] [remote 172.31.0.71:136] [2019-09-30 01:16:56,207] ERROR in app: Exception on / [GET]
[Mon Sep 30 01:16:56.209131 2019] [:error] [pid 2778] [remote 172.31.0.71:136] Traceback (most recent call last):
[Mon Sep 30 01:16:56.209135 2019] [:error] [pid 2778] [remote 172.31.0.71:136]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
[Mon Sep 30 01:16:56.209138 2019] [:error] [pid 2778] [remote 172.31.0.71:136]     response = self.full_dispatch_request()
[Mon Sep 30 01:16:56.209142 2019] [:error] [pid 2778] [remote 172.31.0.71:136]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
[Mon Sep 30 01:16:56.209145 2019] [:error] [pid 2778] [remote 172.31.0.71:136]     rv = self.handle_user_exception(e)
[Mon Sep 30 01:16:56.209148 2019] [:error] [pid 2778] [remote 172.31.0.71:136]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
[Mon Sep 30 01:16:56.209152 2019] [:error] [pid 2778] [remote 172.31.0.71:136]     reraise(exc_type, exc_value, tb)
[Mon Sep 30 01:16:56.209155 2019] [:error] [pid 2778] [remote 172.31.0.71:136]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
[Mon Sep 30 01:16:56.209165 2019] [:error] [pid 2778] [remote 172.31.0.71:136]     raise value
[...]

PR #420, opened 2023-06-30
Update `contrib/azure.py` to enable automatic token refreshes

### LINKED ISSUE #232: Azure provider and auto_refresh_url
Just wondering if there was a reason why the azure provider doesn't seem to have auto_refresh_url configured?

mgd722:
Similarly to the Google behavior (outlined in #143 and #161), Azure will not send a refresh token unless you include the `offline_access` scope. I modeled my changes after `contrib/google.py` to stay consistent, and this update only sets the `auto_refresh_url` if you pass `offline=True` to the blueprint.

Closes #232 and #286 with a built-in solution instead of a login redirect.

### Important note: 
For some reason, [oauthlib raises a full-blown exception](https://github.com/oauthlib/oauthlib/issues/562) for a scope mismatch instead of just a warning. Azure removes the `offline_access` scope, so using automatic refreshes you will always get an error like this:
```
Warning: Scope has changed from "openid offline_access User.Read profile email" to "User.Read openid email profile".
```
The kludgy workaround is to suppress the warning by setting the following environment variable:
```
OAUTHLIB_RELAX_TOKEN_SCOPE=1
```

### DIFF
--- a/flask_dance/contrib/azure.py
+++ b/flask_dance/contrib/azure.py
@@ -11,6 +11,7 @@ def make_azure_blueprint(
     client_secret=None,
     *,
     scope=None,
+    offline=False,
     redirect_url=None,
     redirect_to=None,
     login_url=None,
@@ -34,6 +35,8 @@ def make_azure_blueprint(
         client_id (str): The client ID for your application on Azure AD.
         client_secret (str): The client secret for your application on Azure AD
         scope (str, optional): comma-separated list of scopes for the OAuth token
+        offline (bool): Whether to request offline access for the OAuth token.
+            Defaults to False, but must be True to receive a refresh token.
         redirect_url (str): the URL to redirect to after the authentication
             dance is complete
         redirect_to (str): if ``redirect_url`` is not defined, the name of the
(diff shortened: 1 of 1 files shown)


daenney:
Hiya! Thanks for the PR, I think the change looks good. Two small follow-up things to do before we merge this:

* Add a test to https://github.com/singingwolfboy/flask-dance/blob/main/tests/contrib/test_azure.py to check the new behaviour
* Add a note in https://github.com/singingwolfboy/flask-dance/blob/main/docs/providers.rst about the need to relax the token scope in Azure

mgd722:
Sounds good, those are things I can do, but as a heads up: due to work deadlines and PTO it will likely take until mid/late July.

emer-bestseller:
Are there any updates on this PR? I am facing the same issue, and this seems like the "proper" fix

singingwolfboy:
> Are there any updates on this PR? I am facing the same issue, and this seems like the "proper" fix

It looks like the author of this PR has not had time to do the necessary work. If this is important to you, you could open a new PR that includes these changes, and the extra work required to get them merged.

mgd722:
Looks like https://github.com/singingwolfboy/flask-dance/pull/428/ saw this through better than me, and is merged into prod. Closing.

PR #422, opened 2023-09-12
Enable token refresh for Gitlab

ryan-s:
Gitlab removed support for optionally expiring tokens in Gitlab 15.0.  Adding in the refresh as optional since there may be self hosted instances in the wild older then 15.0.

https://docs.gitlab.com/ee/integration/oauth_provider.html#access-token-expiration

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Changelog
 
 `unreleased`_
 -------------
-nothing yet
+* Added auto refresh support for Gitlab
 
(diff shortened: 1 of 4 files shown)


ryan-s:
Pinned pillow to 9.5 based on https://bugs.launchpad.net/tacker/+bug/2026345

daenney:
> Pinned pillow to 9.5 based on https://bugs.launchpad.net/tacker/+bug/2026345

Ah, nice catch. Thank you!

PR #428, opened 2024-01-11
Enable token refresh for azure

emilhe:
This PR adds the capability of automatic token refresh to the azure bindings.

### DIFF
--- a/flask_dance/contrib/azure.py
+++ b/flask_dance/contrib/azure.py
@@ -11,6 +11,7 @@ def make_azure_blueprint(
     client_secret=None,
     *,
     scope=None,
+    offline=False,
     redirect_url=None,
     redirect_to=None,
     login_url=None,
@@ -34,6 +35,8 @@ def make_azure_blueprint(
         client_id (str): The client ID for your application on Azure AD.
         client_secret (str): The client secret for your application on Azure AD
         scope (str, optional): comma-separated list of scopes for the OAuth token
+        offline (bool): Whether to request offline access for the OAuth token.
+            Defaults to False, but must be True to receive a refresh token.
         redirect_url (str): the URL to redirect to after the authentication
             dance is complete
         redirect_to (str): if ``redirect_url`` is not defined, the name of the
@@ -76,6 +79,15 @@ def make_azure_blueprint(
     :returns: A :doc:`blueprint <flask:blueprints>` to attach to your Flask app.
     """
     scope = scope or ["openid", "email", "profile", "User.Read"]
+    token_url = f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token"
+    authorization_url = (
+        f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize"
+    )
+    auto_refresh_url = None  # see #161
+    if offline:
+        scope = [scope] if not isinstance(scope, list) else scope
+        scope.append("offline_access")
+        auto_refresh_url = token_url
     authorization_url_params = {}
     if login_hint:
         authorization_url_params["login_hint"] = login_hint
@@ -90,8 +102,9 @@ def make_azure_blueprint(
         client_secret=client_secret,
         scope=scope,
         base_url="https://graph.microsoft.com",
-        authorization_url=f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize",
-        token_url=f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token",
+        authorization_url=authorization_url,
+        token_url=token_url,
+        auto_refresh_url=auto_refresh_url,
         redirect_url=redirect_url,
         redirect_to=redirect_to,
         login_url=login_url,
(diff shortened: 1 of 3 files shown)


daenney:
Hiya, thanks for the change! It looks good, but the test file no longer passes the formatting check. All you should have to do is run `black` and it'll correct it. After that I think we can merge it.

flask_dance/contrib/azure.py
```diff
@@ -76,6 +79,15 @@ def make_azure_blueprint(
...
     """
     scope = scope or ["openid", "email", "profile", "User.Read"]
+    token_url = f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token"
+    authorization_url = (
+        f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize"
+    )
+    auto_refresh_url = None  # see #161
+    if offline:
+        scope = [scope] if not isinstance(scope, list) else scope
+        scope.append("offline_access")
```
singingwolfboy, line 89: Modifying the `scope` variable feels like a bad pattern to me. If a developer uses `flask-dance` and specifies the `scope` variable, they expect that the contents of that variable are passed to the OAuth provider as they specified it; I can see this lead to many "WTFs" when debugging.

Assuming that the `offline_access` scope is required to support the token refresh feature, I think it's better to invert the logic here. Rather than this function accepting an `offline` boolean and modifying the scope, this function could not add a new variable, and simply check to see if `"offine_access"` is present in the `scope` variable. If so, the token refresh feature is automatically enabled; if not, then it is not.

What do you think?

singingwolfboy (changes requested):

flask_dance/contrib/azure.py
```diff
@@ -76,6 +79,15 @@ def make_azure_blueprint(
     :returns: A :doc:`blueprint <flask:blueprints>` to attach to your Flask app.
     """
     scope = scope or ["openid", "email", "profile", "User.Read"]
+    token_url = f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token"
+    authorization_url = (
+        f"https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize"
+    )
+    auto_refresh_url = None  # see #161
```
singingwolfboy, line 86: What does `see #161` mean?

emilhe: It's a reference to the issue where the "idea" of setting `auto_refresh_url ` to `None`, if the `offline` option is set to `False`. I guess it might be unnecessary - I can just remove it?

https://github.com/singingwolfboy/flask-dance/pull/161

flask_dance/contrib/azure.py
emilhe: Conceptually, I think I like that idea.  I guess it depends on which weights the most,

1) A common interface, and abstract away technical details
2) A light-weight (technically) transparent interface

If we choose (1), we should stick with the `offline`  keyword (or a similar syntax) as it defines a common way to enable offline access mode across contribs. We thus abstract away the fact that Azure uses a `scope` modification, while e.g. Google uses a `authorization_url_params` modification.

If we choose (2), I believe your proposed syntax is preferred. If case we go down this route, we may consider changing other contribs (e.g. Google) accordingly.

One thing to note though is that even though the scope of the session is modified, the scope of the resulting token is not,

image

I am not sure if this affects your suggestion.

singingwolfboy: > Conceptually, I think I like that idea. I guess it depends on which weights the most,
> 
> 1. A common interface, and abstract away technical details
> 2. A light-weight (technically) transparent interface

Unfortunately, I think that option 1 is not possible. The only common interface between different OAuth providers is OAuth itself -- and more often than not, providers are not even 100% compliant with the OAuth spec! Trying to shoehorn different providers into a common interface is not going to work. Just because Google does it one way, doesn't mean that Microsoft is going to follow suit; in fact, competing companies often try to be intentionally different from each other!

I think making the interface clear & debuggable is a better approach. How would you suggest changing the Google contrib to make it more clear & debuggable?

Also, I was not aware that Azure returns a token with different scopes than the requested scopes -- but if anything, it just provides more evidence that trying to provide a unified interface across providers is not possible. We should not try to obscure these implementation details, particularly since the provider could change their implementation details at any moment.

flask_dance/contrib/azure.py
singingwolfboy: Yeah, I think it can be removed.

flask_dance/contrib/azure.py
emilhe: That's true. While a coherent wrapping would be nice, keeping the library maintained to accomodate the latest changes across contribs is infeasible. I'll change the implementation to match your suggestion.

singingwolfboy:
Could you add something to the docstring of that function, mentioning the automatic token refresh and how to activate it? If you can include a link to the Azure documentation about this topic, that would be even better!

Remember that Sphinx uses this docstring to build the documentation for this project, so people who want to use Flask-Dance will see this, even if they never look at the code!

emilhe:
That's a good idea. I have added a docstring and a link.

singingwolfboy:
This looks great to me! Thanks @emilhe!

PR #433, opened 2024-04-07
fix Discord prompt parameter

owencompher:
the [Discord authentication url](https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-authorization-url-example) expects the prompt argument to be "consent" or "none", by default "consent". If set as `None` in the blueprint constructor, it remains `None` until it is passed to the url, where it is left off entirely. Changing it to `"none"` fixes the issue without the possibility of breaking any other providers.

It looks like this is the only provider with this kind of issue.

### DIFF
--- a/flask_dance/contrib/discord.py
+++ b/flask_dance/contrib/discord.py
@@ -58,7 +58,7 @@ def make_discord_blueprint(
     scope = scope or ["identify"]
     authorization_url_params = {"prompt": "consent"}
     if prompt is None:
-        authorization_url_params["prompt"] = None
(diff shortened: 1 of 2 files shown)


singingwolfboy:
Great! Could you please add a test to cover these cases? (Both "none" and "consent" should be tested.)

owencompher:
I would consider tweaking the docs to remove the ambiguity between the string `"None"` and the builtin `None`. I'm not sure about the best way to do that so I'll leave it up to you

singingwolfboy:
Thanks so much!

